PR #56, opened 2022-11-14
fix: link arrow position

gorgeousvlad:


yuberdysheva:
I noticed that arrows moved to the right and bottom a little

Before:
image

After:
image

yuberdysheva (approved):

PR #61, opened 2022-11-25
fix(ExtendedFeatures): add props

yuberdysheva:


src/blocks/ExtendedFeatures/__stories__/data.json
```diff
@@ -96,18 +80,16 @@
...
-            "theme": "normal"
-          },
-          "links": [
+          "buttons": [
             {
-              "text": "Go",
-              "url": "#",
-              "arrow": true,
-              "theme": "normal"
+              "text": "Button\r",
```
aeksandla, line 85: why is '\r' needed here? and in line 28

yuberdysheva: It's copy paste,  removed

aeksandla (approved):

PR #98, opened 2022-12-27
Feat/tests

niktverd:


src/components/Anchor/__tests__/Anchor.test.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+
+import Anchor from '../Anchor';
+
+const testId = 'anc-hor';
```
gorgeousvlad, line 6: 'anchor'?

src/navigation/components/NavigationItem/NavigationItem.tsx
```diff
@@ -28,7 +28,7 @@ export interface NavigationItemProps {
 
 const Content: React.FC<{text: string; icon?: ImageProps}> = ({text, icon}) => (
     <Fragment>
-        {icon && <Image className={b('icon')} {...icon} />}
+        {icon && typeof icon !== 'string' && <Image className={b('icon')} {...icon} />}
```
gorgeousvlad, line 31: why do we need this fix?

gorgeousvlad (approved):

src/navigation/components/NavigationItem/NavigationItem.tsx
niktverd: Because a new version of typescript requires a destructed item to be an Object. Icon type is complex: one of its type could be a string. String cannot be destructed

src/components/Anchor/__tests__/Anchor.test.tsx
niktverd: fixed

PR #127, opened 2023-02-01
Add media card content

gorgeousvlad:


NikitaCG:
should we add **metaInfo** in `schema.ts` files ?

PR #164, opened 2023-02-25
test(Button): test for Button added

niktverd:


src/components/Button/utils.ts
```diff
@@ -1,5 +1,7 @@
 import {ButtonSize, ButtonView} from '@gravity-ui/uikit';
 
+export const ICON_QA = 'button-icon';
```
yuberdysheva, line 3: Why is ICON_QA here? Maybe it should be replaced to Button.test.tsx?

src/components/Button/__tests__/Button.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+import Button from '../Button';
+
+const qaId = 'button-component';
+
+const buttonProps = {
+    text: 'Button Text',
+    url: 'https://github.com/gravity-ui/',
+    target: '_blank',
+    img: {
+        url: 'https://storage.yandexcloud.net/cloud-www-assets/constructor/storybook/images/img-gray.png',
```
yuberdysheva, line 18: it is image for the button? Why is it not icon?

yuberdysheva (changes requested):

src/components/Button/utils.ts
niktverd: It is used in both test and component files. Let him be here )

src/components/Button/__tests__/Button.test.tsx
niktverd: Sure

yuberdysheva (approved):

PR #166, opened 2023-02-25
test(ButtonTabs): test for ButtonTabs added

niktverd:


src/components/ButtonTabs/__tests__/ButtonTabs.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    test('call onSelectTab', async () => {
+        const user = userEvent.setup();
+        const handleOnClick = jest.fn();
+        render(<ButtonTabs items={items} qa={qaId} onSelectTab={handleOnClick} />);
+
+        const buttons = screen.getAllByRole('button');
+
+        for (let i = 0; i < buttons.length; i++) {
```
gorgeousvlad, line 64: not critical but in previos test you used `forEach` and there is `for` for almost same - mb use on notation?

gorgeousvlad (approved):

src/components/ButtonTabs/__tests__/ButtonTabs.test.tsx
niktverd: good point

PR #169, opened 2023-02-26
test(BackgroundImage): test for BackgroundImage added

niktverd:


src/components/Image/Image.tsx
```diff
@@ -7,11 +7,14 @@ import {isCompressible} from '../../utils/imageCompress';
 import ImageBase from '../ImageBase/ImageBase';
 import i18n from './i18n';
 
+export const qaIdByDefault = 'anchor-component';
```
yuberdysheva, line 10: Is it necessary to add default qa?

niktverd: Actually, yes. I discussed it amje when I was doing uikit test. Sometimes it is too difficult to access tag nested deeply in component. But props you are testing has effect on the tag. And to access it I add default qa value

yuberdysheva (approved):

PR #175, opened 2023-02-28
fix(ExtendedFeatures): label position fixed

niktverd:
https://caniuse.com/?search=column-gap


yuberdysheva:
Label should be placed after the last word immediately. You should check it with designer
image

DaffPunks (approved):

PR #206, opened 2023-03-07
fix: add  map block to text transform blocks

Lakate:


src/text-transform/blocks.ts
```diff
@@ -289,6 +289,13 @@ const config: BlocksConfig = {
             transformer: yfmTransformer,
         },
     ],
+    [BlockType.MapBlock]: [
+        ...blockHeaderTransfomer,
+        {
+            fields: ['title', 'additionalInfo'],
```
yuberdysheva, line 295: We already transform `title` in blockHeaderTransfomer

yuberdysheva (changes requested):

yuberdysheva (approved):

Issue #210, opened 2023-03-09
Wrap Content sub-block buttons with RouterLink


shishovKA:
Content sub-blocks [buttons](https://github.com/gravity-ui/page-constructor/blob/main/src/sub-blocks/Content/Content.tsx#L102) could be wrapped with `RouterLink` component, as its made in [Header](https://github.com/gravity-ui/page-constructor/blob/main/src/blocks/Header/Header.tsx#L143) block.

PR #214, opened 2023-03-11
feat: maintein routerLink in components

niktverd:


niktverd:
https://github.com/gravity-ui/page-constructor/issues/210

src/blocks/Icons/Icons.tsx
```diff
@@ -16,7 +16,7 @@ const Icons = ({title, size = 's', items}: IconsBlockProps) => {
         <div className={b({size})}>
             {title && <BlockHeader className={b('header')} title={title} colSizes={{all: 12}} />}
             {items.map((item) => (
-                <a
+                <a // should we render a Link if it is in LocationContext?
```
gorgeousvlad, line 19: I think yes, we can render Link everywhere, but it may not support properties like `ref`, 'rel' and `onClick`, so it can be a problem in some cases you marked below

PR #215, opened 2023-03-13
feat: add whitelist of image extenstions which can be compressed

niktverd:


src/utils/imageCompress.ts
```diff
@@ -0,0 +1,8 @@
+export enum AvailableForCompressExtention {
```
NikitaCG, line 1: small typos `extenSion`

src/utils/imageCompress.ts
```diff
@@ -0,0 +1,8 @@
...
+    PNG = 'png',
+    JPG = 'jpg',
+    JPEG = 'jpeg',
+}
+
+export const canBeCompressed = (fileName: string) =>
+    Object.keys(AvailableForCompressExtention).find((ext) => fileName.endsWith(`.${ext}`));
```
NikitaCG, line 8: `.some` work similarly like `.find`, but always return `boolean`. If we need boolean in  [disableWebp](https://github.com/gravity-ui/page-constructor/pull/215/files#diff-da73b37ebbe2383290a1d6caf9037c4598e9d500d0632596c8dc877a191ad10fR40) variable always, we need to use `.some`

niktverd: ok

NikitaCG (approved):

src/utils/imageCompress.ts
```diff
@@ -0,0 +1,8 @@
+export enum AvailableForCompressExtension {
+    PNG = 'png',
+    JPG = 'jpg',
+    JPEG = 'jpeg',
+}
+
+export const canBeCompressed = (fileName: string) =>
```
yuberdysheva, line 7: I'm not sure about this name, maybe it can be like `isCompressible`

yuberdysheva (changes requested):

src/utils/imageCompress.ts
niktverd: ok

PR #225, opened 2023-03-21
feat: refactor text-transform functions

NikitaCG:


niktverd (approved):

src/text-transform/common.ts
```diff
@@ -0,0 +1,42 @@
...
+    items.map((item) => {
+        if (!item) {
+            return item;
+        } else if (typeof item === 'string') {
+            return transformer(item);
+        } else {
+            return {
+                ...item,
+                ...fields.reduce<ComplexItem>((acc, fieldName) => {
+                    const result = {...acc};
```
gorgeousvlad, line 22: do we really need to copy acc here?

src/text-transform/transformers.ts
```diff
@@ -0,0 +1,94 @@
...
+
+import {ConstructorBlock} from '../models/constructor';
+import {fullTransform} from './utils';
+
+import {Block} from '../models';
+
+import {Lang} from '../utils/configure';
+
+import {Transformer} from './common';
+
```
gorgeousvlad, line 11: why do we need an empty line after each `import` statement here?

src/text-transform/transformers.ts
```diff
@@ -0,0 +1,94 @@
...
+    options: {
+        lang: Lang;
+        customConfig?: {};
+    };
+};
+
+function addRandomOrder(block: ConstructorBlock) {
+    if (block) {
+        if ('randomOrder' in block && block.randomOrder && 'children' in block && block.children) {
+            block.children = _.shuffle(block.children as ConstructorBlock[]);
```
gorgeousvlad, line 28: not to mutate incoming `block` param:
```
return {
  ...block,
  children: _.shuffle(block.children as ConstructorBlock[]);
}
```

src/text-transform/transformers.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+}
+
+function transformBlocks(blocks: ConstructorBlock[], lang: Lang, customConfig = {}) {
+    const fullConfig = {...config, ...customConfig};
+
+    const transformedBlocks = _.cloneDeep(blocks);
```
gorgeousvlad, line 36: blocks are not transformed yet at this line

gorgeousvlad (changes requested):

src/text-transform/common.ts
NikitaCG: Initially, I did not rewrite the logic, I left the previous version. I've corrected it now

src/text-transform/transformers.ts
NikitaCG: fix name

src/text-transform/transformers.ts
NikitaCG: remove `addRandomOrder' because it's useless as a separate function, and insert this logic inside transformfunc

src/text-transform/transformers.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    return clonedBlocks.map((block) => transformBlock(lang, fullConfig, block));
+}
+
+function transformBlock(lang: Lang, blocksConfig: BlocksConfig, block: ConstructorBlock) {
+    const blockConfig = blocksConfig[block.type];
+
+    if (block) {
+        if ('randomOrder' in block && block.randomOrder && 'children' in block && block.children) {
+            // eslint-disable-next-line no-not-accumulator-reassign/no-not-accumulator-reassign
```
gorgeousvlad, line 33: тут точно нужен этот коммент?

gorgeousvlad (approved):

PR #237, opened 2023-03-27
fix(Tabs): wrap tabs on desktop

DC-RomanKarpov:


styles/mixins.scss
```diff
@@ -518,18 +518,14 @@ unpredictable css rules order in build */
...
     &_centered {
-        display: flex;
         justify-content: center;
-        flex-wrap: wrap;
     }
 
     @media (max-width: map-get($gridBreakpoints, 'md')) {
-        display: flex;
         flex-wrap: nowrap;
         justify-content: flex-start;
```
yuberdysheva, line 530: Is it supposed to be here without display flex?

yuberdysheva (changes requested):

styles/mixins.scss
DC-RomanKarpov: The `display: flex` property will be inherited from the base selector. So I think yes, it can be here without additional property setup.

yuberdysheva: justify-content: flex-start; already exists in the base selector then)

yuberdysheva (changes requested):

styles/mixins.scss
DC-RomanKarpov: But it can be overriden by the &_centered's property assignment.

yuberdysheva (approved):

PR #254, opened 2023-03-29
feat: dropdown a ReactPlayer version

niktverd:


src/components/ReactPlayer/ReactPlayer.tsx
```diff
@@ -32,6 +32,7 @@ import {useAnalytics} from '../../hooks';
 import {PlayVideo} from '../../icons';
 
 import './ReactPlayer.scss';
+import {checkYoutubeVideos} from './utils';
```
yuberdysheva, line 35: this line should be before styles

src/components/ReactPlayer/__stories__/data.json
```diff
@@ -1,7 +1,7 @@
 {
   "default": {
     "content": {
-      "src": "https://storage.yandexcloud.net/cloud-www-assets/constructor/storybook/images/video_8-12_white.mp4",
+      "src": ["https://youtube.com/live/XNRVbfUMkS0"],
```
yuberdysheva, line 4: Why we changed default config? There is a config for youtube

yuberdysheva (changes requested):

aeksandla (approved):

PR #262, opened 2023-03-30
fix(ReactPlayer): transform youtube url for single video src

niktverd:


niktverd:
tested locally with @1.26.2-alpha.0

aeksandla (approved):

PR #303, opened 2023-04-13
feat: drop rules overriding

niktverd:


src/blocks/FilterBlock/FilterBlock.tsx
```diff
@@ -6,9 +6,9 @@ import {ConstructorBlocks} from '../../containers/PageConstructor/components/Con
 import {Col, Row} from '../../grid';
 import {BlockType, ConstructorItem, FilterBlockProps, FilterItem} from '../../models';
 import {block} from '../../utils';
-import i18n from './i18n';
 
 import './FilterBlock.scss';
+import i18n from './i18n';
```
yuberdysheva, line 11: oh no.. is this from new eslint-rules? It can be right

src/blocks/Header/Header.tsx
```diff
@@ -9,9 +9,9 @@ import {ThemeValueContext} from '../../context/theme/ThemeValueContext';
 import {Col, Grid, Row} from '../../grid';
 import {ClassNameProps, HeaderBlockBackground, HeaderBlockProps, WithChildren} from '../../models';
 import {block, getThemedValue} from '../../utils';
-import {getImageSize, getTitleSizes, titleWithImageSizes} from './utils';
 
 import './Header.scss';
+import {getImageSize, getTitleSizes, titleWithImageSizes} from './utils';
```
yuberdysheva, line 14: this is also some interesting things, maybe we have to discuss it in opensorce chat?

yuberdysheva (changes requested):

src/text-transform/utils.ts
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable no-param-reassign */
```
yuberdysheva, line 1: We added this rule. Why we are turning it off here?

src/utils/scroll.ts
```diff
@@ -58,7 +58,7 @@ function getChangeScrollFunction(
     isReverse: boolean,
 ): ChangeScrollFunction {
     return (element: HTMLDivElement, scrollPosition: number, startPosition: number) => {
-        // eslint-disable-next-line no-not-accumulator-reassign/no-not-accumulator-reassign
+        // eslint-disable-next-line no-param-reassign, no-not-accumulator-reassign/no-not-accumulator-reassign
```
yuberdysheva, line 61: We added this rule. Why we are turning it off here?

niktverd: Because of the way we call the function in setTimeout we have difficulties to return value. That is the reason we need to modify input argument

gorgeousvlad (approved):

src/text-transform/utils.ts
niktverd: I have created task. We will investigate this problem and fix it

PR #304, opened 2023-04-13
fix(BasicCard): schema validation

yuberdysheva:


src/schema/validators/common.ts
```diff
@@ -567,7 +567,7 @@ export const MapProps = {
 export const CardBase = {
     border: {
         type: 'string',
-        enum: ['border', 'shadow', 'none'],
+        enum: ['line', 'shadow', 'none'],
```
gorgeousvlad, line 570: would old card with border `border` also work?

gorgeousvlad (approved):

src/schema/validators/common.ts
yuberdysheva: It was always a property `line`. I think it's a mistake. You can check properties of CardBase

PR #308, opened 2023-04-18
feat(QuestionBlock): add schema.org FAQ microdata

niktverd:


src/blocks/Questions/Questions.tsx
```diff
@@ -28,7 +38,7 @@ const QuestionsBlock = (props: QuestionsProps) => {
     };
 
     return (
-        <div className={b()}>
+        <div className={b()} itemScope itemType={FaqMicrodataValues.PageType}>
```
yuberdysheva, line 41: What does it means this props `itemScope`?

niktverd: As I understand, passing itemscope to element initiates a context creation for new meta variable.
In the issue, you have highlighted, new variable describing FAQ block were initiated
Below, every time we meet a new Question block we have to initiate a new variable with a context in order to path content related features (Question name, Answer text)

itemscope is a kind of a border among these meta variables

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemscope

yuberdysheva (approved):

PR #329, opened 2023-05-04
feat(Navigation): add iconSize for navigation icons, update margin between icon and text

aeksandla:
…tween icon and text

### DIFF
--- a/src/navigation/components/Header/Header.tsx
+++ b/src/navigation/components/Header/Header.tsx
@@ -1,4 +1,4 @@
-import React, {MouseEvent, useCallback, useState} from 'react';
+import React, {MouseEvent, useCallback, useMemo, useState} from 'react';
 
 import _ from 'lodash';
 
@@ -6,7 +6,12 @@ import Control from '../../../components/Control/Control';
 import OutsideClick from '../../../components/OutsideClick/OutsideClick';
 import {Col, Grid, Row} from '../../../grid';
 import {NavigationClose, NavigationOpen} from '../../../icons';
-import {HeaderData, ThemedNavigationLogoData} from '../../../models';
+import {
+    HeaderData,
+    NavigationItemBase,
+    NavigationItemModel,
+    ThemedNavigationLogoData,
+} from '../../../models';
 import {block} from '../../../utils';
 import {ItemColumnName} from '../../constants';
 import Logo from '../Logo/Logo';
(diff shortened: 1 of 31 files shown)


src/navigation/components/Header/Header.tsx
```diff
@@ -52,11 +57,32 @@ const MobileMenuButton: React.FC<MobileMenuButtonProps> = ({
...
 export const Header: React.FC<HeaderProps> = ({data, logo}) => {
-    const {leftItems, rightItems} = data;
+    const {leftItems, rightItems, iconSize} = data;
     const [isSidebarOpened, setIsSidebarOpened] = useState(false);
     const [activeItemId, setactiveItemId] = useState<string | undefined>(undefined);
 
+    const getNavigationItemWithIconSize = useCallback(
+        (item: NavigationItemModel) => {
+            if (!(iconSizeKey in item)) {
+                return {...item, iconSize};
```
yuberdysheva, line 70: What happens if iconSize is not set? Maybe you should add the default value

yuberdysheva (changes requested):

src/navigation/components/Header/Header.tsx
aeksandla: If iconSize is not set, then no styles will be applied to the image.
seeIconSizeStyle: https://github.com/gravity-ui/page-constructor/pull/329/files#diff-01cc3be42ecfb8f27579002a8ce8d95d1aa53569e00008dd1510400ed43425e7R18

yuberdysheva: Maybe we should set some default size?

src/navigation/components/NavigationItem/components/ContentWrapper/ContentWrapper.scss
```diff
@@ -7,7 +7,7 @@ $block: '.#{$ns}content-wrapper';
         display: flex;
         width: 20px;
         height: 20px;
```
yuberdysheva, line 9: can we delete this styles with default iconSize?

yuberdysheva (approved):

src/navigation/components/NavigationItem/components/ContentWrapper/ContentWrapper.scss
aeksandla: of course

yuberdysheva (approved):

PR #346, opened 2023-05-22
feat(Button): add prop width

yuberdysheva:


gorgeousvlad (dismissed):

src/components/Button/Button.tsx
```diff
@@ -59,8 +60,8 @@ const Button = (props: ButtonProps) => {
             : {url: img, position: defaultImgPosition};
 
     const buttonClass = img
-        ? b({position: buttonImg.position, size, theme}, className)
-        : b({size, theme}, className);
+        ? b({position: buttonImg.position, size, theme, width}, className)
+        : b({size, theme, width}, className);
```
gorgeousvlad, line 64: mb there we already need variable like 
`const buttonModifiers = {size, theme, width}`

gorgeousvlad (approved):

PR #363, opened 2023-05-31
feat(navigation): add prop withBorder by default

yuberdysheva:


src/navigation/__stories__/PageConstructor.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {NAVIGATION} from '../../demo/constants';
+
+import data from './data.json';
+
+export default {
+    title: `${NAVIGATION}/PageConstructor`,
+    component: PageConstructor,
+} as Meta;
+
+const DefaultTemplate: Story<PageConstructorProps> = (args) => <PageConstructor {...args} />;
```
aeksandla, line 15: If this story is about navigation, then maybe you need to show the navigation's props only, and make the rest of the page unchanged?

`const DefaultTemplate: Story<NavigatinProps> = (args) => <PageConstructor {..._.omit(data, 'navigation')} {...args} />;`


for example:
```
DefaultNavigation.args = data.navigation as NavigationProps;
```

aeksandla (approved):

PR #364, opened 2023-05-31
fix: filter block nesting

gorgeousvlad:


src/blocks/FilterBlock/FilterBlock.tsx
```diff
@@ -78,7 +73,13 @@ const FilterBlock: React.FC<FilterBlockProps> = ({
                 </Row>
             )}
             <Row className={b('block-container')}>
-                <ConstructorBlocks items={container} />
+                <CardLayoutBlock title="" colSizes={colSizes}>
```
yuberdysheva, line 76: Why is `title` empty?

gorgeousvlad: > Why is `title` empty?

it was same before, didn't change anything here

yuberdysheva: Can you make `title` as not necessary in CardLayoutBlock? It looks not so good

yuberdysheva (approved):

PR #369, opened 2023-06-01
feat: update header and navigation

yuberdysheva:
Add new stories for navigation:
https://preview.gravity-ui.com/page-constructor/369/?path=/story/navigation-navigation--default-navigation

### DIFF
--- a/src/navigation/components/Header/Header.tsx
+++ b/src/navigation/components/Header/Header.tsx
@@ -1,5 +1,7 @@
 import React, {MouseEvent, useCallback, useEffect, useMemo, useState} from 'react';
 
+import _ from 'lodash';
+
 import Control from '../../../components/Control/Control';
 import OutsideClick from '../../../components/OutsideClick/OutsideClick';
 import {Col, Grid, Row} from '../../../grid';
--- a/src/navigation/components/Header/Header.tsx
+++ b/src/navigation/components/Header/Header.tsx
@@ -1,4 +1,4 @@
-import React, {MouseEvent, useCallback, useMemo, useState} from 'react';
+import React, {MouseEvent, useCallback, useEffect, useMemo, useState} from 'react';
 
 import Control from '../../../components/Control/Control';
 import OutsideClick from '../../../components/OutsideClick/OutsideClick';
@@ -68,9 +68,10 @@ const isDropdownItem = (item: NavigationItemModel): item is NavigationDropdownIt
     item.type === NavigationItemType.Dropdown;
 
 export const Header: React.FC<HeaderProps> = ({data, logo}) => {
-    const {leftItems, rightItems, iconSize = 20} = data;
+    const {leftItems, rightItems, iconSize = 20, withBorder = false} = data;
     const [isSidebarOpened, setIsSidebarOpened] = useState(false);
     const [activeItemId, setactiveItemId] = useState<string | undefined>(undefined);
+    const [withHeaderBorder, setWithHeaderBorder] = useState(withBorder);
 
     const getNavigationItemWithIconSize = useCallback(
         (item: NavigationItemModel) => {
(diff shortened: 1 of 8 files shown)


src/navigation/components/Header/Header.tsx
```diff
@@ -114,8 +117,21 @@ export const Header: React.FC<HeaderProps> = ({data, logo}) => {
...
+    useEffect(() => {
+        const handleScroll = () => {
+            if (window.scrollY > 0 && !withBorder) {
+                setWithHeaderBorder(true);
+            } else if (window.scrollY === 0 && !withBorder) {
+                setWithHeaderBorder(false);
+            }
+        };
+
+        window.addEventListener('scroll', _.debounce(handleScroll, 5), {passive: true});
```
gorgeousvlad, line 129: mb set time to 100?

src/navigation/components/Header/Header.tsx
```diff
@@ -114,8 +117,21 @@ export const Header: React.FC<HeaderProps> = ({data, logo}) => {
         setIsSidebarOpened(false);
     }, []);
 
+    useEffect(() => {
+        const handleScroll = () => {
```
gorgeousvlad, line 121: seems like we should add this handler only if !withBorder is true

src/navigation/components/Header/Header.tsx
```diff
@@ -68,9 +70,10 @@ const isDropdownItem = (item: NavigationItemModel): item is NavigationDropdownIt
     item.type === NavigationItemType.Dropdown;
 
 export const Header: React.FC<HeaderProps> = ({data, logo}) => {
-    const {leftItems, rightItems, iconSize = 20} = data;
+    const {leftItems, rightItems, iconSize = 20, withBorder = false} = data;
     const [isSidebarOpened, setIsSidebarOpened] = useState(false);
     const [activeItemId, setactiveItemId] = useState<string | undefined>(undefined);
+    const [withHeaderBorder, setWithHeaderBorder] = useState(withBorder);
```
gorgeousvlad, line 76: what is the difference between withBorder and withBorder?

src/navigation/components/Header/Header.tsx
```diff
@@ -114,8 +117,21 @@ export const Header: React.FC<HeaderProps> = ({data, logo}) => {
         setIsSidebarOpened(false);
     }, []);
 
+    useEffect(() => {
+        const handleScroll = () => {
+            if (window.scrollY > 0 && !withBorder) {
+                setWithHeaderBorder(true);
```
gorgeousvlad, line 123: why do we set withHeaderBorder when withBorder is false?

aeksandla (dismissed):

src/navigation/components/Header/Header.tsx
yuberdysheva: - We pass the prop `withBorder` which can show or hide border when `scrollY = 0`. 
- We always have to show border when `scrollY > 0`. 
- I chose not to rewrite the original prop to hide/show border but create a new one `withHeaderBorder` and change this state if it's necessary. I rename variable `withHeaderBorder` to `showBorder` is that better?

src/navigation/components/Header/Header.tsx
yuberdysheva: Because `withBorder = false` needs only for `scrollY = 0`. If we have  `scrollY > 0` we have to show border all the time

src/navigation/components/Header/Header.tsx
yuberdysheva: if I set 100 there is big delay on scroll before border shows up. I set 20

src/navigation/components/Header/Header.tsx
gorgeousvlad: ` If we have scrollY > 0 we have to show border all the time`
then should this condition be like `if(window.scrollY > 0)`?

gorgeousvlad (dismissed):

src/navigation/components/Header/Header.tsx
yuberdysheva: It's not necessary. If `withBorder = true` we don't have to  change anything.
`const [showBorder, setShowBorder] = useState(withBorder);`
`showBorder` is constantly `true`

yuberdysheva: I think we can check `showBorder ` instead of `withBorder ` if you think so

gorgeousvlad (approved):

PR #386, opened 2023-06-06
fix: constructor items inner data structure

gorgeousvlad:


src/customization/BlockDecoration.tsx
```diff
@@ -1,27 +1,24 @@
...
-export const BlockDecoration = (props: PropsWithChildren<BlockDecorationProps>) => {
-    const blockContenxtId = getBlockIndexFromId(useContext(BlockIdContext));
+export const BlockDecoration = ({
+    id,
+    children: blockChildren,
+}: PropsWithChildren<BlockDecorationProps>) => {
     const {headerBlockTypes} = useContext(InnerContext);
-    const blockId = props.id || blockContenxtId;
-    const isHeader = Boolean(props.id && headerBlockTypes.includes(props.id));
+    const isHeader = Boolean(typeof id === 'string' && headerBlockTypes.includes(id));
```
yuberdysheva, line 11: Why don't we render header if id === number?

yuberdysheva (approved):

src/customization/BlockDecoration.tsx
gorgeousvlad: it is just flag passed for editor it is not used it constructor itself

PR #397, opened 2023-06-13
feat: add exceptions for typograf rules

NikitaCG:


src/text-transform/utils.ts
```diff
@@ -28,7 +28,12 @@ export const DEFAULT_ALLOWED_TAGS = [
 ];
 export const typografConfig = {
     enabled: ['common/nbsp/afterNumber', 'common/nbsp/afterParagraphMark'],
-    disabled: ['common/symbols/cf', 'ru/other/phone-number'],
+    disabled: [
+        'common/symbols/cf',
+        'ru/other/phone-number',
+        'common/space/afterColon',
+        'common/space/afterSemicolon',
```
aeksandla, line 35: why do we remove the space after the semicolon and the formatting of the number?

NikitaCG: - after semicolon we have same problem, and content want to fix it
- it's history, maybe maintainers khow

aeksandla (approved):

gorgeousvlad (approved):

yuberdysheva (approved):

PR #415, opened 2023-06-19
ci: add alpha workflow

yuberdysheva:


.github/workflows/release-alpha.yml
```diff
@@ -2,20 +2,48 @@ name: Release alpha version
 
 on:
   workflow_dispatch:
+    inputs:
+      manualPackageVersion:
+        type: string
+        required: false
+        description: 'If you have any troubles with release you can set version of package manually, e.g. 3.0.0-alpha.0'
```
gorgeousvlad, line 9: what kind of troubles could it be?)

.github/workflows/release-alpha.yml
```diff
@@ -2,20 +2,48 @@ name: Release alpha version
...
-      - name: Extract branch name
-        shell: bash
-        run: |
-          echo "branch_name=${GITHUB_REF#refs/heads/}"
-        id: extract_branch
-      - uses: gravity-ui/release-action@v1
+      - run: |
+          if [ "${{ github.event.inputs.manualPackageVersion }}" != "" ]; then
+            if [[ "${{ github.event.inputs.manualPackageVersion }}" == *"alpha"* ]]; then
+              echo "manualPackageVersion is assigned correctly!"
```
gorgeousvlad, line 18: mb we should only message if name is incorrect?

.github/workflows/release-alpha.yml
```diff
@@ -2,20 +2,48 @@ name: Release alpha version
...
-          npm-dist-tag: alpha
-          default-branch: ${{ env.branch_name }}
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+        shell: bash
+      - run: npm test
+        shell: bash
+      - name: Bump and commit version
+        run: |
+          git config --local user.email "111915794+gravity-ui-bot@users.noreply.github.com"
```
gorgeousvlad, line 35: are't we have this email in ENV variables?

.github/workflows/release-alpha.yml
yuberdysheva: E.g: version already exists

gorgeousvlad (approved):

PR #416, opened 2023-06-20
fix: semantic markup

VitaliiDC8:


src/blocks/Header/Header.scss
```diff
@@ -77,6 +77,8 @@ $backgroundWidth: 1440px;
     &__description {
         margin-top: $indentXS;
         font-weight: normal;
+        font-size: 0.83em;
```
gorgeousvlad, line 80: we usually use font sizes from defined set https://preview.gravity-ui.com/uikit/?path=/story/typography--common

src/blocks/Header/Header.tsx
```diff
@@ -130,12 +130,12 @@ export const HeaderBlock = (props: WithChildren<HeaderBlockFullProps>) => {
                                         <HTML>{title}</HTML>
                                     </h1>
                                     {description && (
-                                        <h5 className={b('description')}>
+                                        <div className={b('description')}>
```
gorgeousvlad, line 133: why?

src/blocks/Slider/Arrow/Arrow.tsx
```diff
@@ -19,14 +19,14 @@ export interface ArrowProps {
 const Arrow = ({type, handleClick, className, size = 16}: ArrowProps & ClassNameProps) => (
     <div className={b({type}, className)}>
         <button className={b('button')} onClick={() => handleClick && handleClick(type)}>
-            <div className={b('icon-wrapper')}>
+            <span className={b('icon-wrapper')}>
```
gorgeousvlad, line 22: also why?

src/blocks/Slider/Slider.tsx
```diff
@@ -224,24 +224,28 @@ export const SliderBlock = (props: WithChildren<SliderProps>) => {
...
         return (
             slidesCountByBreakpoint > 1 && (
-                <div
-                    className={b('bar')}
-                    style={{
-                        left: barPosition,
-                        width: barWidth,
-                    }}
-                />
+                <li>
```
gorgeousvlad, line 227: why slider by is <li>?

src/blocks/Slider/Slider.tsx
```diff
@@ -224,24 +224,28 @@ export const SliderBlock = (props: WithChildren<SliderProps>) => {
...
     };
 
     const renderDot = (index: number) => {
         return (
-            <div
-                key={index}
-                className={b('dot', {active: index === currentIndex})}
-                onClick={() => handleDotClick(index)}
-            />
+            <li>
```
gorgeousvlad, line 242: we should check didn't it add additional <li> styles

src/navigation/components/NavigationItem/components/NavigationDropdown/NavigationDropdown.tsx
```diff
@@ -16,7 +16,7 @@ const TOGGLE_ARROW_SIZE = 12;
 type NavigationDropdownProps = NavigationItemProps & DropdownItemData;
 
 export const NavigationDropdown = React.forwardRef<HTMLElement, NavigationDropdownProps>(
-    ({text, icon, isOpened, className, iconSize, ...props}, ref) => {
+    ({text, icon, isOpened, className, iconSize, type: _, ...props}, ref) => {
```
gorgeousvlad, line 19: ??

src/utils/blocks.ts
```diff
@@ -15,6 +15,14 @@ export function getHeaderTag(size: TextSize) {
     }
 }
 
+export function hasBlockTag(content: string): boolean {
+    const blockElementRegex =
+        /<(div|p|h[1-6]|address|article|aside|blockquote|canvas|dd|dl|dt|fieldset|figcaption|footer|header|hr|main|nav|section|video|pre|figure|form|pre|ol|ul|li|table|thead|tbody|tfoot|tr|th|td)\b[^>]*>/gi;
```
gorgeousvlad, line 20: let's make an aray with tags like
`BLOCK_ELEMENTS = ['div', 'p',...]`
and then `BLOCK_ELEMENTS.join('|')`

src/navigation/components/NavigationItem/components/NavigationDropdown/NavigationDropdown.tsx
VitaliiDC8: Attribute type not allowed on element [span](https://html.spec.whatwg.org/multipage/#the-span-element) at this point. span tag does not have 'type' attribute and validator.w3.org informs us about this. So I cannot see technical reason to set it

src/blocks/Slider/Slider.tsx
VitaliiDC8: validator.w3.org informs us that 'ul' tag should have only 'li' tag as a child

src/blocks/Header/Header.tsx
VitaliiDC8: because 'h' tags are using for headers. but description is not header and this is error for html semantic. I have changed it to paragraph.

src/blocks/Companies/Companies.tsx
```diff
@@ -17,7 +17,7 @@ export const CompaniesBlock = ({title, images, animated}: CompaniesBlockProps) =
     return (
         <AnimateBlock className={b()} offset={150} animate={animated}>
             <div className={b('content')}>
-                <h2 className={b('title')}>{title}</h2>
+                {title && <h2 className={b('title')}>{title}</h2>}
```
yuberdysheva, line 20: `CompaniesBlockProps` has prop `title` as obligatory

src/blocks/Slider/Slider.tsx
```diff
@@ -224,24 +224,24 @@ export const SliderBlock = (props: WithChildren<SliderProps>) => {
 
         return (
             slidesCountByBreakpoint > 1 && (
-                <div
+                <li
```
yuberdysheva, line 227: Maybe wrapper of this element is semantically not `<ul>`?
I saw there only dots and arrows

src/navigation/components/NavigationItem/components/NavigationDropdown/NavigationDropdown.tsx
yuberdysheva: Is `type` used anywhere?
Maybe it can be `div` instead of `span`?

src/utils/blocks.ts
```diff
@@ -1,6 +1,43 @@
 import {Block, CustomConfig, PCShareSocialNetwork, TextSize} from '../models';
 import {ConstructorBlock} from '../models/constructor';
 
+const BLOCK_ELEMENTS = [
```
yuberdysheva, line 4: With which element we could render span in HTML?

yuberdysheva (changes requested):

src/navigation/components/NavigationItem/components/NavigationDropdown/NavigationDropdown.tsx
VitaliiDC8: how I see we don't read elements to get his attribute. 'Type' is used somewhere upper in component. I have decided not to change to div, because for my case it does not matter span or div in this component.

src/blocks/Slider/Slider.tsx
VitaliiDC8: I see that dots in ul

src/utils/blocks.ts
yuberdysheva: But why you chose all these to render `div`?

VitaliiDC8: In HTML component we can choose div or span tag. And in this case I check if child component is block element, if it is I take div as parent. Otherwise we take default 'span' tag.

yuberdysheva (approved):

PR #420, opened 2023-06-22
feat: add yfm transform for editor demo

gorgeousvlad:


package.json
```diff
@@ -89,7 +89,7 @@
     "react-spring": "^9.3.0",
     "react-transition-group": "^4.4.2",
     "react-waypoint": "^10.1.0",
-    "sanitize-html": "^2.6.1",
+    "sanitize-html": "2.6.1",
```
yuberdysheva, line 92: why?

gorgeousvlad: 2.8 and higher version of `sanitize-html` breaks dev build, I fixed working version.

yuberdysheva (approved):

PR #424, opened 2023-06-22
Editor fixes pt 2

gorgeousvlad:


src/editor/containers/Form/Form.tsx
```diff
@@ -57,25 +57,28 @@ export const Form = memo(({content, onChange, activeBlockIndex, onSelect, spec}:
...
-                                        ...blocks.slice(index + 1),
-                                    ],
-                                });
-                            }}
-                            onSelect={() => onSelect(index)}
-                        />
-                    ))}
+                    {blocks.map(
+                        (blockData, index) =>
+                            blocksSpec[blockData.type] && (
```
NikitaCG, line 62: if there is no specification, will undefined be returned ? can we use the ternary operator to return null for security?

NikitaCG (changes requested):

NikitaCG (approved):

yuberdysheva (approved):

PR #426, opened 2023-06-23
Editor style fixes

gorgeousvlad:


src/editor/components/BlockForm/BlockForm.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../../../styles/variables.scss';
+@import '../../../../styles/mixins.scss';
+@import '../../styles/mixins.scss';
+@import '../../styles/variables.scss';
```
yuberdysheva, line 4: Are these imports necessary here?

PR #427, opened 2023-06-25
feat: add transform when variables handling

gorgeousvlad:


src/text-transform/filter.ts
```diff
@@ -0,0 +1,100 @@
...
+                    });
+                }
+            }
+        }
+
+        return result;
+    }, []);
+}
+
+function checkWhenCondition(whenValue: WhenValue, vars: ContentVariables) {
```
NikitaCG, line 57: its Function Declaration, but I think good practice is rearrange this func before `filterItems`

NikitaCG (dismissed):

src/text-transform/filter.ts
gorgeousvlad: It always was a controversial point where to place function dependencies before or after it (even in our team) and I think there is no strict rule for it

yuberdysheva (dismissed):

yuberdysheva (approved):

PR #433, opened 2023-06-28
fix: regex to match child content on block tags

VitaliiDC8:


src/utils/blocks.ts
```diff
@@ -53,10 +55,10 @@ export function getHeaderTag(size: TextSize) {
 }
 
 export function hasBlockTag(content: string): boolean {
-    const blockElementRegex = `/<(${BLOCK_ELEMENTS.join('|')})\b[^>]*>/gi`;
-    const blockTags = content.match(blockElementRegex);
+    const regex = new RegExp(BLOCK_ELEMENTS_REGEX, 'g');
+    const isBlock = regex.test(content);
```
gorgeousvlad, line 59: `return regex.test(content);`

gorgeousvlad (approved):

PR #448, opened 2023-07-24
feat: YAML tab (read-only) in editor

Kyzyl-ool:


src/editor/components/YamlEditor/YamlEditor.tsx
```diff
@@ -0,0 +1,32 @@
...
+import yaml from 'js-yaml';
+import MonacoEditor from 'react-monaco-editor';
+
+import {PageContent} from '../../../models';
+
+interface YamlEditorProps {
+    content: PageContent;
+}
+
+export const YamlEditor = ({content}: YamlEditorProps) => {
```
NikitaCG, line 12: return immediately

NikitaCG (changes requested):

PR #456, opened 2023-07-25
feat: change content layout block props

Kyzyl-ool:
- Removed `properties` prop from `ContentLayoutBlock`

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -15,4 +15,4 @@ CONTRIBUTING.md
 
 # npm files
 package.json
-
+/widget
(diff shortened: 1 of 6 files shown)


gorgeousvlad (approved):

.prettierignore
```diff
@@ -15,4 +15,4 @@ CONTRIBUTING.md
 
 # npm files
 package.json
-
+/widget
```
Lunatic174, line 18: Doesn't look like npm file)

Kyzyl-ool: agree) that folder was duplicated, removed

PR #459, opened 2023-07-27
Uikit5

gorgeousvlad:

PR #475, opened 2023-08-02
Uikit5 (#459)

gorgeousvlad:
* feat: update to uikit-5


src/components/Button/Button.tsx
```diff
@@ -1,6 +1,7 @@
 import React, {useCallback, useContext} from 'react';
 
-import {Button as CommonButton, Icon, Platform, StoreBadge} from '@gravity-ui/uikit';
+import {StoreBadge} from '@gravity-ui/components';
```
Kyzyl-ool, line 3: `StoreBadge` seems not to have `lang` prop and you should use `href` instead of `url`

Kyzyl-ool (changes requested):

src/components/Button/Button.tsx
Kyzyl-ool: Oops, just saw the fix here https://github.com/gravity-ui/page-constructor/pull/476

yuberdysheva (approved):

Kyzyl-ool (approved):

PR #481, opened 2023-08-07
feat: add an icon position on the left side of a basic-card

niktverd:


src/sub-blocks/BasicCard/BasicCard.scss
```diff
@@ -4,15 +4,42 @@
 $block: '.#{$ns}basic-card';
 
 #{$block} {
+    $iconSizeL: 32px;
```
gorgeousvlad, line 7: why do we have L and S sizes, but don't have m?

src/sub-blocks/BasicCard/__stories__/BasicCard.stories.tsx
```diff
@@ -36,6 +37,20 @@ const WithIconTemplate: StoryFn<BasicCardProps> = (args) => (
     </div>
 );
 
+const WithIconLeftTemplate: StoryFn<BasicCardProps> = (args) => (
```
yuberdysheva, line 40: I think that would be nice to do one tab `with Icon` and show there default position and left position

src/sub-blocks/BasicCard/BasicCard.scss
yuberdysheva, line 7: Why it's name is size L? 
If the size depends on position maybe we can name it something like `iconSizePositionLeft`?

gorgeousvlad (changes requested):

src/sub-blocks/BasicCard/BasicCard.tsx
```diff
@@ -5,29 +5,47 @@ import CardBase from '../../components/CardBase/CardBase';
...
 const BasicCard = (props: BasicCardProps) => {
-    const {title, text, icon, additionalInfo, links, buttons, ...cardParams} = props;
+    const {
+        title,
+        text,
+        icon,
+        additionalInfo,
+        links,
+        buttons,
+        iconPosition = IconPosition,
```
yuberdysheva, line 23:
```suggestion
        iconPosition = IconPosition.Top,
```

yuberdysheva (changes requested):

src/sub-blocks/BasicCard/BasicCard.scss
niktverd: Good point

src/sub-blocks/BasicCard/__stories__/BasicCard.stories.tsx
niktverd: Ok

src/sub-blocks/BasicCard/BasicCard.scss
niktverd: I've changed it

src/sub-blocks/BasicCard/__stories__/BasicCard.stories.tsx
```diff
@@ -23,28 +23,14 @@ const DefaultTemplate: StoryFn<BasicCardProps> = (args) => (
     </div>
 );
 
-const WithIconTopTemplate: StoryFn<BasicCardProps> = (args) => (
+const WithIconTemplate: StoryFn<BasicCardProps> = (args) => (
```
yuberdysheva: Can we left here only two cards with specific headers?
you can see the example of borders 
https://preview.gravity-ui.com/page-constructor/481/?path=/story/components-cards-basiccard--with-border

niktverd: agree

gorgeousvlad (approved):

yuberdysheva (approved):

PR #500, opened 2023-08-14
fix(Icons): make icon without href no clickable

aeksandla:


src/blocks/Icons/Icons.tsx
```diff
@@ -8,24 +8,35 @@ import {block, getLinkProps} from '../../utils';
 import './Icons.scss';
 
 const b = block('icons-block');
-
 const Icons = ({title, size = 's', items}: IconsBlockProps) => {
     const {hostname} = useContext(LocationContext);
 
+    const itemContent = (item: IconsBlockProps['items'][number]) => (
```
yuberdysheva, line 14: useMemo? or it can be located outside of `Icons`

src/blocks/Icons/Icons.tsx
```diff
@@ -8,24 +8,35 @@ import {block, getLinkProps} from '../../utils';
...
-                    key={item.url}
-                    href={item.url}
-                    {...getLinkProps(item.url, hostname)}
-                >
-                    <Image className={b('image')} src={item.src} />
-                    <p className={b('text')}>{item.text}</p>
-                </a>
-            ))}
+            {items.map((item) =>
+                item.url ? (
```
yuberdysheva, line 25: we can create here variable and call the function `itemContent(item)` once

yuberdysheva (approved):

Issue #488, opened 2023-08-08
Questions block


Kyzyl-ool:
1. Each question element should be accessible via TAB navigation and expandable on Enter
1. Each question element should have `aria-expanded` attribute set correctly

PR #505, opened 2023-08-16
feat: Questions block keyboard access improved

Kyzyl-ool:
Closes https://github.com/gravity-ui/page-constructor/issues/488


src/blocks/Questions/Questions.tsx
```diff
@@ -22,6 +24,65 @@ const FaqMicrodataValues = {
     AnswerTextProp: 'text',
 } as const;
 
+const QuestionBlockItem = ({
```
yuberdysheva, line 27: Can we create sub-directory for QuestionBlockItem?

Kyzyl-ool: of course. done)

PR #538, opened 2023-09-04
fix: remove unused youtube js on the first loading

VitaliiDC8:


src/components/VideoBlock/VideoBlock.tsx
```diff
@@ -115,12 +115,14 @@ const VideoBlock = (props: VideoBlockProps) => {
...
         }
 
-        const fullSrc = `${src}?${getPageSearchParams(attributes || {})}`;
-
         if (ref.current && !iframeRef.current) {
             const iframe = document.createElement('iframe');
             iframe.id = fullId;
-            iframe.src = fullSrc;
+
+            if (fullscreen && !previewImg) {
```
gorgeousvlad, line 122: why should we only use src when it is fullscreen, how about non-fullscreen cases?

VitaliiDC8: below in code we have check to show preview image and this is combined with fullSchreen. But ok, I guess you are right, we can remove this from condition and check only !previewImg

gorgeousvlad (approved):

PR #540, opened 2023-09-04
feat: migrate to `@bem-react/classname`

pavel-klimov:


src/components/YFMWrapper/YFMWrapper.tsx
```diff
@@ -1,15 +1,13 @@
-/* we won't use 'pc' class prefix here to let you opportunity to define yfm styles
```
yuberdysheva, line 1: Why you removed a comment?

.eslintignore
```diff
@@ -1,6 +1,5 @@
 node_modules
 
-/eslint-plugin-local
```
yuberdysheva, line 3: Why you removed it? It was needed for bem-cn-lite?

pavel-klimov: It's local plugin that didn't connect to project and it may checking that `bem-cn-lite` didn't import directly. I think, It's not need any more

src/components/YFMWrapper/YFMWrapper.tsx
pavel-klimov: Apparently it's my mistake. I return it

package.json
```diff
@@ -120,7 +120,7 @@
     "@gravity-ui/prettier-config": "^1.0.1",
     "@gravity-ui/stylelint-config": "^1.0.0",
     "@gravity-ui/tsconfig": "^1.0.0",
-    "@gravity-ui/uikit": "^5.4.1",
+    "@gravity-ui/uikit": "^5.9.1",
```
gorgeousvlad, line 123: why?

gorgeousvlad (approved):

package.json
pavel-klimov: It's remove `bem-cn-lite` from dependencies. Old version worked with `bem-cn-lite`. I changed it. New version work with `@bem-react/classname`.

PR #549, opened 2023-09-08
feat: ability to add titles for the links

Kyzyl-ool:
- Added `urlTitle` everywhere where links (`<a>` tags) were used.

This feature can be used to provide both visual and accessible warning that the link will open in a new tab


src/navigation/components/NavigationItem/components/NavigationLink/NavigationLink.tsx
```diff
@@ -31,14 +31,20 @@ export const NavigationLink: React.FC<NavigationLinkProps> = (props) => {
 
     if (linkExtraProps?.target || !Link) {
         return (
-            <a href={url} title={text} className={classes} {...rest} {...linkExtraProps}>
+            <a
+                href={url}
+                title={urlTitle || text}
```
yuberdysheva, line 36: why is there a fallback to the text in this place but not in all places?

src/navigation/components/SocialIcon/SocialIcon.tsx
```diff
@@ -13,15 +13,15 @@ export interface NavigationSocialItemOwnProps extends NavigationSocialItem {
...
+const SocialIcon: React.FC<NavigationSocialItemOwnProps> = ({icon, url, className, urlTitle}) => {
     const iconData = getMediaImage(icon);
     const {alt} = iconData;
 
     return (
         <a
             href={url}
             aria-label={alt}
-            title={alt}
+            title={urlTitle || alt}
```
yuberdysheva, line 24: Can `alt` of icon be the title of a link?

src/navigation/components/NavigationItem/components/NavigationLink/NavigationLink.tsx
Kyzyl-ool: For backward compatibility I left fallback here. But I didn't do one on other places to avoid unexpected tooltips. Moreover, I expect that the link text and link title will not be the same

src/navigation/components/SocialIcon/SocialIcon.tsx
Kyzyl-ool: Yes, but the accessible name will read:
1) link text
2) image alt
3) link title

Is icon's `alt` is the same as the link title,  the screenreader will read two sentences twice, it is not really good. So I think they should differ

yuberdysheva (approved):

PR #554, opened 2023-09-12
feat: add style schema to ImageObjectProps

niktverd:


src/components/Image/schema.ts
```diff
@@ -13,6 +13,15 @@ const ImageBase = {
     },
 };
 
+const StyleBase = {
+    type: 'object',
+    additionalProperties: false,
+    required: [],
+    properties: {
+        backgroundColor: {type: 'string'},
```
yuberdysheva, line 21: Here could be any css property, not only backgroundColor
maybe `string`

niktverd: added a few widespread options: `width`, `height`, `color`

yuberdysheva (approved):

PR #622, opened 2023-10-06
add new view for Table and fix design bugs

Lakate:
1) add new table view 
image

2) fix phone fields in  hubspot form 
in prod: 
image

in beta:
image

3) fix double background-color styles in BackgroundImage component


src/sub-blocks/HubspotForm/HubspotForm.scss
```diff
@@ -110,6 +110,16 @@ $block: '.#{$ns}hubspot-form';
                     border: 1px solid var(--g-color-line-generic-active);
                 }
             }
+
+            .hs-fieldtype-intl-phone.hs-input .hs-input {
+                float: none;
+                /* stylelint-disable-next-line declaration-no-important */
+                width: 100% !important;
```
Lakate, line 117: hubspot script set important for these elements, so i need to also use important

yuberdysheva: Could you please add this comment into the file?

src/components/BackgroundImage/BackgroundImage.tsx
```diff
@@ -15,7 +15,11 @@ const BackgroundImage = (props: WithChildren<BackgroundImageProps>) => {
     const qaAttributes = getQaAttrubutes(qa || qaIdByDefault);
 
     return (
-        <div className={b(null, className)} style={style} data-qa={qa || qaIdByDefault}>
+        <div
+            className={b(null, className)}
+            style={{...style, backgroundColor: 'transparent'}}
```
yuberdysheva, line 20: why did you set backgroundColor here but not in css-file?

src/components/Table/Table.tsx
```diff
@@ -18,7 +21,7 @@ const b = block('table');
 
 export default class Table extends React.Component<TableProps & ClassNameProps> {
     render() {
-        const {content, legend, marker = 'disk', className, caption} = this.props;
+        const {content, legend, hideLegend, marker = 'disk', className, caption} = this.props;
```
yuberdysheva, line 24: Maybe we can check if legend exists? In what case we can pass a legend and hide it?

yuberdysheva:
Could you please add a new type of table to the storybook? Didn't find it there
https://preview.gravity-ui.com/page-constructor/622/?path=/story/blocks-table--default

yuberdysheva (changes requested):

src/components/BackgroundImage/BackgroundImage.tsx
Lakate: because styles from html have higher specific than styles in css

src/components/Table/Table.tsx
Lakate: Now we have this part of code https://github.com/gravity-ui/page-constructor/blob/main/src/components/Table/Table.tsx#L44-L48

And if change logic here, it will break all created tables

src/sub-blocks/HubspotForm/HubspotForm.scss
Lakate: Done

yuberdysheva (approved):

PR #655, opened 2023-10-20
feat: alt attributes and titles

Kyzyl-ool:
- alt attributes for Quote block and Author avatar
- share links with labels and titles


src/components/Image/i18n/ru.json
```diff
@@ -1,3 +1 @@
-{
-  "img-alt": ""
-}
+{}
```
yuberdysheva, line 1: Do we need these empty files?

src/schema/validators/common.ts
```diff
@@ -321,10 +321,7 @@ export const authorItem = {
             type: 'string',
             contentType: 'text',
         },
-        avatar: {
-            type: 'string',
-            pattern: imageUrlPattern,
-        },
+        avatar: ImageProps,
```
yuberdysheva, line 324: Could it be a breaking change?

Kyzyl-ool: no, it actually expands the use cases as ImageProps contains previous schema:
Screenshot 2023-10-23 at 12 18 25

src/components/Image/i18n/ru.json
Kyzyl-ool: No, we can delete them

yuberdysheva (approved):

PR #668, opened 2023-10-27
fix: raise up title item arrow to 1 px

Kyzyl-ool:


niktverd (approved):

src/components/Title/__stories__/Title.stories.tsx
```diff
@@ -57,6 +94,10 @@ CustomTitle.args = {
 Sizes.args = {
     ...DefaultArgs,
 } as TitleProps;
+SizesWithLinks.args = {
+    ...DefaultArgs,
+    title: data.titleLink.content.title,
```
gorgeousvlad, line 99: do we use this, check pls

Kyzyl-ool: yep, `titleLink` is used in `TitleLink` story

gorgeousvlad (approved):

PR #672, opened 2023-11-01
fix: aligning left quote

Kyzyl-ool:


src/sub-blocks/Quote/Quote.scss
```diff
@@ -7,7 +7,6 @@ $block: '.#{$ns}quote';
     &_#{$name} {
         &::before {
             position: absolute;
```
gorgeousvlad, line 9: seems like for other quote type it is still actual

Kyzyl-ool: here it already applied for the left quote only, but for all `quoteType`'s. No need to align for the right one

gorgeousvlad (approved):

PR #674, opened 2023-11-03
feat: a11y descriptions and labels

Kyzyl-ool:


src/components/BackLink/BackLink.tsx
```diff
@@ -9,14 +9,15 @@ import {DefaultEventNames, Tabbable} from '../../models';
...
-export interface BackLinkProps extends Tabbable {
+export interface BackLinkProps<T = HTMLElement> extends Tabbable {
     url: string;
     title: ReactNode;
     theme?: Theme;
     size?: ButtonSize;
     className?: string;
     shouldHandleBackAction?: boolean;
     onClick?: () => void;
+    extraProps?: HTMLProps<T>;
```
gorgeousvlad, line 20: maybe `htmlProps` or `htmlElementProps` would be more obvious?

src/sub-blocks/BasicCard/BasicCard.tsx
```diff
@@ -37,7 +45,9 @@ const BasicCard = (props: BasicCardProps) => {
                     )}
                     <Content
                         title={title}
+                        titleId={titleId}
```
gorgeousvlad, line 48: is it possible to create this ids inside the `Content` component? Do we need some manual id?

gorgeousvlad (changes requested):

src/components/BackLink/BackLink.tsx
Kyzyl-ool: The name `extraProps` is the commonly used name in the uikit, so I think this variant is appropriate here

src/sub-blocks/BasicCard/BasicCard.tsx
Kyzyl-ool: Yes, but in this case it will be impossible to label and describe `CardBase`, so one had to support both types: ids from props and generated inside Content by default

gorgeousvlad (approved):

PR #688, opened 2023-11-10
test(Companies): test for Companies block added

niktverd:


src/blocks/Companies/__tests__/Companies.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+import {getQaAttrubutes} from '../../../utils';
+import Companies from '../Companies';
+
+const companies: CompaniesBlockProps = {
+    title: 'Companies Title',
+    animated: true,
+    images: {
+        light: {
+            desktop:
+                'https://storage.yandexcloud.net/cloud-www-assets/constructor/storybook/images/logo-svg_12-12_desktop_light.svg',
```
gorgeousvlad, line 17: It is maybe not necessary now, but I get task to move all images that constructor depends on inside project. It means we are trying to avoid adding images from storage.yandexcloud.net now

gorgeousvlad (approved):

yuberdysheva (approved):

PR #693, opened 2023-11-14
feat: add standalone navigation

gorgeousvlad:


src/navigation/components/NavigationItem/NavigationItem.tsx
```diff
@@ -21,7 +22,7 @@ const NavigationItem: React.FC<NavigationItemProps> = ({
     ...props
 }: NavigationItemProps) => {
     const {type = NavigationItemType.Link} = data;
-    const {navItemMap} = useContext(InnerContext);
+    const navItemMap = useNavItemMap();
```
yuberdysheva, line 25: Just a suggestion:
```suggestion
    const navItemMap = useNavigationItemMap();
```

src/navigation/components/Standalone/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import RootCn from '../../../components/RootCn';
+
+import Navigation, {NavigationProps} from './../../components/Navigation/Navigation';
+
+const Standalone = (props: NavigationProps) => (
```
yuberdysheva, line 7: Why is const exists and its name Standalone?

gorgeousvlad: I called it so because you can import and use this Navigation component regardless of constructor itself

yuberdysheva: Can we add to readme the example of how to import navigation and use it with theme?

yuberdysheva (approved):

PR #696, opened 2023-11-15
feat: add description field to icons block

VitaliiDC8:


yuberdysheva:
Can you add please:
1. transform support of title and description. Example: https://github.com/gravity-ui/page-constructor/blob/38d5b884cd222d42e6f512e38f913fc8602a7ab1/src/text-transform/config.ts#L363
2. File Icon.mdx with properties of block. (not necessary but would be great!)
Thank you.

yuberdysheva (approved):

PR #697, opened 2023-11-15
feat: add PriceCard

qradle-yndx:


src/components/ContentList/ContentList.tsx
```diff
@@ -25,25 +25,34 @@ function getHeadingLevel(size: ContentSize) {
...
+    const renderIcon = (icon: ContentItemProps['icon']): JSX.Element => {
+        const iconThemed = getThemedValue(icon, theme);
+        if (isIconSvg(iconThemed)) {
+            const Icon = iconThemed;
+            return <div>{<Icon className={b('icon')} />}</div>;
+        }
+        const iconData = getMediaImage(iconThemed);
+        return <Image {...iconData} className={b('icon')} qa={qaAttributes.image} />;
+    };
+
```
NikitaCG, line 48: can we moved this in a separate  file component ? together with this utility `isIconSvg`

src/sub-blocks/PriceCard/PriceCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+                                        icon: Check,
+                                        text,
+                                    }))}
+                                    size="s"
+                                />
+                            </div>
+                        ) : null}
+                    </div>
+                    {button && <Button className={b('button')} {...button} />}
+                    {link && <LinkBlock className={b('link')} textSize="m" {...link} />}
```
NikitaCG, line 59: if there is an empty object, will it work correctly?

src/sub-blocks/PriceCard/PriceCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import {Button, CardBase, ContentList, Link as LinkBlock} from '../../components';
+import {PriceCardProps} from '../../models';
+import {block} from '../../utils';
+
+import './PriceCard.scss';
+
+const b = block('price-card');
+
+const PriceCard = (props: PriceCardProps) => {
```
NikitaCG, line 13: maybe rewrite like this :
```
({border, ...}) => (<CardBase>....</CardBase>)
```

src/models/constructor-items/sub-blocks.ts
```diff
@@ -134,6 +135,18 @@ export interface BannerCardProps {
...
+export interface PriceCardProps extends CardBaseProps, Pick<ContentBlockProps, 'theme'> {
+    title: string;
+    price: string;
+    pricePeriod?: string;
+    priceDetails?: string;
+    description?: string;
+    button?: ButtonProps;
+    link?: LinkProps;
+    backgroundColor?: string;
+    list?: string[];
```
NikitaCG, line 147: what kind of list ? maybe we should give a more semantic name, for example `priceList` ?

src/sub-blocks/PriceCard/PriceCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+                            </div>
+                            {priceDetails && (
+                                <div className={b('price-details')}>{priceDetails}</div>
+                            )}
+                        </div>
+                        <div>{description}</div>
+                        {list?.length ? (
+                            <div className={b('list')}>
+                                <ContentList
+                                    list={list.map((text) => ({
```
NikitaCG, line 49: what kind of text?) if we rename list in priceList or something else, maybe we need transform `text -> price` ? for a better understanding of what the essence is here))

NikitaCG (changes requested):

src/components/ContentList/ContentList.tsx
qradle-yndx: It's possible. But it seems like that renderIcon strongly relates to parent component through theme and qaAttributes params. @yuberdysheva What do you think? Is it worth it?

src/sub-blocks/PriceCard/PriceCard.tsx
qradle-yndx: No it won't. @yuberdysheva Should i fix this in src/sub-blocks/BackgroundCard/BackgroundCard.tsx too?
What would be the expected behaviour? (Don't show these elements until they get all required props, do they)

src/sub-blocks/PriceCard/PriceCard.tsx
qradle-yndx: I can see that other cards have this kinda syntax too. Let's keep it.

src/sub-blocks/PriceCard/PriceCard.tsx
qradle-yndx: who knows?) I would propose to name it benefitsList.

qradle-yndx: so `text -> price` would become to `text -> benefit` and altogether:
```
<ContentList
  list={benefitsList.map((benefit) => ({
    icon: Check,
    text: benefit,
  }))}
  size="s"
/>
```

src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -0,0 +1,121 @@
...
+#{$block} {
+    position: relative;
+
+    &__bg {
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+    }
```
yuberdysheva, line 15: what is it for? we don't use image for background. can we set color to the main cards-component?

src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -0,0 +1,121 @@
...
+
+    &__content {
+        height: 100%;
+        display: flex;
+        flex-direction: column;
+        position: relative;
+
+        &_theme {
+            &_dark {
+                color: var(--g-color-text-light-primary);
```
yuberdysheva, line 25: lets add here 
--g-color-line-focus: var(--pc-color-line-focus-dark);

src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -0,0 +1,121 @@
...
+                #{$block}__info a {
+                    text-decoration: underline;
+
+                    &:hover {
+                        color: var(--g-color-text-light-secondary);
+                    }
+                }
+
+                #{$block}__button {
+                    text-decoration: none;
```
yuberdysheva, line 41: why is it here?

src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -0,0 +1,121 @@
...
+
+    &__main {
+        flex: 1;
+        min-height: 0;
+        display: flex;
+        flex-direction: column;
+        position: relative;
+    }
+
+    &__media {
```
yuberdysheva, line 99: what is media here?

src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -0,0 +1,121 @@
...
+        position: absolute;
+        top: 0;
+        left: 0;
+        bottom: 0;
+        right: 0;
+        border-radius: $borderRadius;
+    }
+
+    &__image {
```
yuberdysheva, line 108: I din't see any images in storybook

src/sub-blocks/PriceCard/PriceCard.tsx
yuberdysheva, line 59: LinkBlock accepts array. Maybe we also should give this opportunity?

yuberdysheva (changes requested):

yuberdysheva:
Let's add:
1. Validation scheme
2. text transform like here https://github.com/gravity-ui/page-constructor/blob/963e29be6a53d7213b4428dfd17f11ab7f6ca0d0/src/text-transform/config.ts#L164

src/sub-blocks/PriceCard/PriceCard.tsx
qradle-yndx: > if there is an empty object, will it work correctly?

I've just realized that case with empty object is not possible. `interface LinkProps` has required `url` prop so it can't be set from the code. The other scenarios would be handled by validator.
There's just one situation with setting empty object on the storybook interface manually but it can be done for any other field either.

src/sub-blocks/PriceCard/PriceCard.scss
qradle-yndx: Replaced div with `BackgroundImage` component according to src/sub-blocks/BackgroundCard/BackgroundCard.tsx and renamed this `&__bg` to `&__background`

src/components/ContentList/ContentListItemIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    if (isIconSvg(iconThemed)) {
+        const Icon = iconThemed;
+        return (
+            <div>
+                <Icon className={className} />
+            </div>
+        );
+    }
+    const iconData = getMediaImage(iconThemed);
```
yuberdysheva, line 29: Can you move it to  other constants on line 19?

src/sub-blocks/PriceCard/schema.ts
```diff
@@ -0,0 +1,47 @@
...
+                type: 'string',
+            },
+            priceDetails: {
+                type: 'string',
+            },
+            description: {
+                type: 'string',
+            },
+            button: ButtonBlock,
+            link: LinkProps,
```
yuberdysheva, line 35:
```suggestion
            buttons: ButtonBlock[],
            links: LinkProps[],
```

src/components/ContentList/ContentListItemIcon.tsx
qradle-yndx: > Can you move it to other constants on line 19?

getImageMedia works only with `image: ImageProps`. iconThemed could be SVGIcon on line 19.

yuberdysheva (approved):

NikitaCG (approved):

PR #698, opened 2023-11-15
feat: add iframe to media

yuberdysheva:


NikitaCG (approved):

src/models/constructor-items/common.ts
```diff
@@ -338,6 +344,20 @@ export interface DataLensObjectProps {
     theme: 'dark' | 'light';
 }
 
+export interface IframeProps {
```
gorgeousvlad, line 347: which of this props we suppose to get from content configs?

src/components/Media/Iframe/Iframe.scss
```diff
@@ -0,0 +1,26 @@
...
+            border-radius: 0;
+        }
+    }
+
+    #{$block}__iframe {
+        border-radius: 0;
+    }
+
+    &__item {
+        width: calc(100% + 4px);
```
gorgeousvlad, line 22: could you please use variables describing why there are 4px indents and -2px margin?

src/models/constructor-items/common.ts
yuberdysheva: I thought that it could be all of this props

gorgeousvlad (approved):

PR #711, opened 2023-11-24
fix: image slider and background

yuberdysheva:


src/containers/PageConstructor/PageConstructor.tsx
```diff
@@ -99,7 +99,7 @@ export const Constructor = (props: PageConstructorProps) => {
         <InnerContext.Provider value={context}>
             <RootCn>
                 <div className={b('wrapper')}>
-                    {themedBackground && (
+                    {Object.keys(themedBackground).length && (
```
gorgeousvlad, line 102: there would be an error if we don't have `themedBackground` value

gorgeousvlad (changes requested):

src/containers/PageConstructor/PageConstructor.tsx
yuberdysheva: I've checked, if background = {} as a default, then themedBackground also {}

NikitaCG (approved):

src/containers/PageConstructor/PageConstructor.tsx
gorgeousvlad: > I've checked, if background = {} as a default, then themedBackground is also {}

why can't we just delete it's default value?

gorgeousvlad (approved):

PR #713, opened 2023-11-27
fix: add transfrom title in html for content item

VitaliiDC8:


src/components/ContentList/ContentList.scss
```diff
@@ -16,6 +16,11 @@ $marginIconSizeL: 1px;
             @include text-size(header-1);
 
             margin: 0 0 $indentXXXS;
+
+            .yfm {
```
yuberdysheva, line 20: you have to add it in size_s too

yuberdysheva (changes requested):

src/components/ContentList/ContentList.scss
VitaliiDC8: added

yuberdysheva: Can we set this styles in one place. I see copy-paste in size l and s

yuberdysheva (approved):

PR #715, opened 2023-11-27
feat: added col size for header in icons block

VitaliiDC8:


src/blocks/Icons/__stories__/Icons.stories.tsx
```diff
@@ -30,9 +30,18 @@ const SizeTemplate: StoryFn<IconsBlockModel> = (args) => (
     </Fragment>
 );
 
+const HeaderColSizeTemplate: StoryFn<IconsBlockModel> = (args) => (
+    <Fragment>
+        <DefaultTemplate {...args} title="ColSize 12" />
```
yuberdysheva, line 35: Can we use here different way, because you can't see all examples in the page in this way and it is not obcious that you can scroll

Example: https://github.com/gravity-ui/page-constructor/blob/af496089edd2bcec2e4f89680001ca0acdc2f828/src/blocks/CardLayout/__stories__/CardLayout.stories.tsx#L33

yuberdysheva (approved):

PR #721, opened 2023-11-28
feat(Navigation): provide dropdown props to custom components

DC-RomanKarpov:


src/navigation/components/NavigationItem/NavigationItem.tsx
```diff
@@ -14,6 +14,13 @@ const b = block('navigation-item');
 
 const ANALYTICS_ID = 'navigation';
 
+const navigationItemTypeValues = Object.values(NavigationItemType);
+const directNavigationItemTypeValues = navigationItemTypeValues;
+directNavigationItemTypeValues.splice(
+    directNavigationItemTypeValues.indexOf(NavigationItemType.Dropdown),
+    1,
+);
+
```
yuberdysheva, line 23: Suggest:
```suggestion
const directNavigationItemTypeValues = navigationItemTypeValues.splice(
    navigationItemTypeValues.indexOf(NavigationItemType.Dropdown),
    1,
);
```

Can you explain what is going on here with splice?

DC-RomanKarpov: The idea here is to get all default `NavigationItemType`s  values except `NavigationItemType.Dropdown`. Because for these values later we'll exclude some props for underlying components.

gorgeousvlad (approved):

src/navigation/__stories__/CustomComponent/CustomComponent.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const CustomComponent: React.FC<DCDropdownNavigationItemProps> = (props) => {
+    const {onClick, isActive, menuLayout} = props;
+    return (
+        <div className={b({active: isActive})} onClick={onClick}>
+            <span className={b('text')}>Custom Item</span>
+            {' ('}
+            {`${menuLayout}`}
+            {isActive && ' - active'}
+            {')'}
```
gorgeousvlad, line 23: `(${menuLayout}) ${isActive ? '- active' : ''}`

PR #746, opened 2023-12-05
fix: add theme to price card validator schema

qradle-yndx:


src/sub-blocks/PriceCard/schema.ts
```diff
@@ -1,11 +1,9 @@
...
-    CardBase,
-    LinkProps,
-    MediaProps,
-} from '../../schema/validators/common';
+import pick from 'lodash/pick';
+
+import {BaseProps, ButtonBlock, CardBase, LinkProps} from '../../schema/validators/common';
+import {ContentBase} from '../Content/schema';
+
+const PriceCardContentProps = pick(ContentBase, ['theme']);
```
gorgeousvlad, line 6: is there only one property we pick? mb jus add like
`
...
theme: ContentBase.theme
...
`?

gorgeousvlad (approved):

PR #753, opened 2023-12-13
Iframe refactoring

niktverd:


src/components/Media/Iframe/Iframe.tsx
```diff
@@ -11,21 +13,91 @@ const b = block('media-component-iframe');
...
+    const handleMessage = useCallback(
+        ({data}: MessageEvent) => {
+            if (height !== 'auto' && typeof height === 'number' && iframeRef.current) {
+                iframeRef.current.height = `${height}px`;
+                return;
+            }
+
+            try {
+                const parsed = JSON.parse(data);
+                const iframeHeight = parsed['iframe-height'];
```
gorgeousvlad, line 64: Where does this message format come from? I think we are trying to keep Iframe component abstract enough

src/components/Media/Iframe/Iframe.tsx
```diff
@@ -11,21 +13,91 @@ const b = block('media-component-iframe');
...
 const Iframe = (props: MediaComponentIframeProps) => {
     const {iframe, margins = true} = props;
-    const {height = 400, src, width, name, title} = iframe;
+    const {height = 400, src, width = '100%', name, title, justifyContent = 'center'} = iframe;
+
+    const formContainerRef = useRef<HTMLDivElement>(null);
+    const iframeRef = useRef<HTMLIFrameElement>();
+    const [iframeId] = useState(uuidv4());
+
+    const updateFormIframe = useCallback(
```
gorgeousvlad, line 22: why do we have form in Iframe component? I can't get the relation, seems like there should be no hard connection between forms and this iframe component

gorgeousvlad (changes requested):

src/components/Media/Iframe/Iframe.tsx
niktverd: Copy + Paste effect. Fixed

src/components/Media/Iframe/Iframe.tsx
niktverd: Copy + Paste effect. Fixed

src/components/Media/Iframe/Iframe.tsx
```diff
@@ -11,21 +13,91 @@ const b = block('media-component-iframe');
 
 const Iframe = (props: MediaComponentIframeProps) => {
     const {iframe, margins = true} = props;
-    const {height = 400, src, width, name, title} = iframe;
+    const {height = 400, src, width = '100%', name, title, justifyContent = 'center'} = iframe;
+
+    const formContainerRef = useRef<HTMLDivElement>(null);
+    const iframeRef = useRef<HTMLIFrameElement>();
+    const [iframeId] = useState(uuidv4());
```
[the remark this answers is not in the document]
niktverd: Get rid of any hooks

src/components/Media/Iframe/Iframe.tsx
niktverd: An iframe is a kind of uniq component. You have no idea, that the content of it can be. That is the reason why we are getting data from inside of an iframe to adjust component size to its height.

Moreover, we already use this approach here https://github.com/gravity-ui/page-constructor/blob/main/src/components/YandexForm/YandexForm.tsx#L117

gorgeousvlad (approved):

PR #754, opened 2023-12-13
feat: add background for card-layout-block

qradle-yndx:


src/blocks/CardLayout/CardLayout.scss
```diff
@@ -8,5 +8,28 @@ $block: '.#{$ns}card-layout-block';
         margin-top: $indentSM;
     }
 
+    &__content {
+        position: relative;
+
+        &_with-background {
+            padding: 8px 32px 48px;
```
gorgeousvlad, line 15: this indents should be from variables

src/blocks/CardLayout/CardLayout.tsx
```diff
@@ -29,19 +29,32 @@ const CardLayout: React.FC<CardLayoutBlockProps> = ({
...
-                <Col key={index} sizes={colSizes} className={b('item')}>
-                    {child}
-                </Col>
-            ))}
-        </Row>
-    </AnimateBlock>
-);
+    background,
+}) => {
+    const withBackground = Boolean(
```
gorgeousvlad, line 34: why not just `const withBackground = Boolean(background)` ?

gorgeousvlad (changes requested):

src/blocks/CardLayout/CardLayout.tsx
qradle-yndx: Theoretically it is possible to pass an empty object to `background` prop which would do nothing inside the `BackgroundImage` rendering but it adds additional indents. 
I added the same logic in the `src/blocks/Form/Form.tsx` as well.
Should I fix it also?

gorgeousvlad (approved):

gorgeousvlad (approved):

PR #767, opened 2024-01-11
fix: card layout block background

qradle-yndx:
- Add border prop for card layout same as in the CardBase.
- Fix card layout schema validation for background

### DIFF
--- a/src/blocks/CardLayout/schema.ts
+++ b/src/blocks/CardLayout/schema.ts
@@ -16,6 +16,10 @@ export const CardLayoutProps = {
         ...BlockHeaderProps,
         colSizes: containerSizesObject,
         background: BackgroundImageProps,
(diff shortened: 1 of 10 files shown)


src/blocks/CardLayout/schema.ts
```diff
@@ -15,7 +15,11 @@ export const CardLayoutProps = {
         ...AnimatableProps,
         ...BlockHeaderProps,
         colSizes: containerSizesObject,
-        background: ImageObjectProps,
+        background: BackgroundImageProps,
+        backgroundBorder: {
```
gorgeousvlad, line 19: could we add it inside `background` property?

gorgeousvlad (changes requested):

gorgeousvlad (approved):

PR #788, opened 2024-01-29
feat: content finalization

lesterrry:


gorgeousvlad:
@lesterrry hi! could you tell please what is the purpose of `FinalizeContent` feature addition?
Currently we are transforming all of our content on server, do you need an opportunity to do it on client?

PR #804, opened 2024-02-07
feat: use i18n instance from uikit

NikitaCG:


src/components/Map/GoogleMap.tsx
```diff
@@ -1,13 +1,13 @@
 import React, {useContext, useEffect, useMemo, useRef, useState} from 'react';
 
+import {Lang} from '@gravity-ui/uikit';
```
korvin89, line 3: U can use `'${Lang}'` type below (use backticks instead of single quotes). It is inconvenient to set such properties because of importing enums

korvin89 (approved):

gorgeousvlad (approved):

PR #807, opened 2024-02-09
feat: add image card sub-block

qradle-yndx:


src/sub-blocks/ImageCard/ImageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </div>
+        );
+    };
+
+    return (
+        <div
+            className={b({border, 'with-content': hasContent, direction})}
+            style={{backgroundColor}}
+        >
+            {direction === ImageCardDirection.Direct && (
```
NikitaCG, line 54: I think we need to change the direction with style, not with code.

`flex-direction` rule in container

src/sub-blocks/ImageCard/ImageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        if (!hasContent) {
+            return null;
+        }
+        return (
+            <div className={b('content')}>
+                <Content title={title} text={text} colSizes={{all: 12, md: 12}} size="s" />
+            </div>
+        );
+    };
+    const renderImage = () => {
```
NikitaCG, line 37: and when we change directions with style we dont need this functions, we'll render this code immediately in return()

src/models/constructor-items/sub-blocks.ts
```diff
@@ -57,6 +58,17 @@ export interface IconWrapperProps {
     icon?: PositionedIcon;
 }
 
+export enum ImageCardMargins {
+    None = 'none',
```
NikitaCG, line 62: do we really need the `none` parameter? We can just have a default state (now is `none`) and change it somehow when there is a demand for `s` or `m`

NikitaCG (changes requested):

NikitaCG (approved):

PR #808, opened 2024-02-12
feat!: version 5

NikitaCG:
* feat: use i18n instance from uikit


stepastepastepa:
I conducted a design review, everything is visually ok https://preview.gravity-ui.com/page-constructor/808/?path=/docs/documentation-blocks--docs

src/components/Button/Button.tsx
```diff
@@ -74,7 +69,7 @@ const Button = (props: ButtonProps) => {
     if (theme === 'app-store' || theme === 'google-play') {
         const platform = theme === 'app-store' ? Platform.IOS : Platform.ANDROID;
 
-        return <StoreBadge className={buttonClass} platform={platform} href={url} />;
+        return <StoreBadge className={buttonClass} platform={platform} href={url as string} />;
```
gorgeousvlad, line 72: why did typecast appear?

src/components/ReactPlayer/ReactPlayer.scss
```diff
@@ -51,7 +51,7 @@ $block: '.#{$ns}ReactPlayer';
     }
 
     &__icon {
-        margin-left: 6px;
+        margin-left: 1px;
```
gorgeousvlad, line 54: why do we include this changes to major?

src/text-transform/transformers.ts
```diff
@@ -17,7 +17,7 @@ export type ContentTransformerProps = {
         blocks?: ConstructorBlock[];
     };
     options: {
-        lang: Lang;
+        lang: `${Lang}`;
```
gorgeousvlad, line 20: could we define this lang type once in models like `type Lang = `${Lang}`?

gorgeousvlad (approved):

src/text-transform/transformers.ts
NikitaCG: I think this is not necessary because we will add another level of abstraction and we will have to go to the place of the announcement to understand what exactly Lang is used, dev experience will deteriorate

src/components/Button/Button.tsx
NikitaCG: `url: string | undefined` , but href now needed only `string`

PR #817, opened 2024-02-14
feat(Components): add className props to be able to configure appearance in clients

DC-RomanKarpov:


src/sub-blocks/Content/Content.tsx
```diff
@@ -43,23 +43,36 @@ function getButtonSize(size: ContentSize) {
     }
 }
 
-export type ContentProps = ContentBlockProps & ClassNameProps & QAProps;
+export type ContentProps = ContentBlockProps &
+    ClassNameProps &
+    QAProps & {
+        titleClassName?: string;
```
gorgeousvlad, line 49: for what purpose we need all of this customisations, mb it is easier to create own block, than redefine everyting
anyway in this case we should create only one object property for all of this classes and remove it from top level properties

src/components/IconWrapper/IconWrapper.scss
```diff
@@ -2,41 +2,42 @@
...
-        &_left {
-            @include add-specificity(&) {
+            #{$contentElement} {
                 flex: 1 0 0;
             }
         }
     }
 }
+
+#{$iconElement} {
```
gorgeousvlad, line 32: why was it moved to top level?

gorgeousvlad (changes requested):

src/components/IconWrapper/IconWrapper.scss
DC-RomanKarpov: class names for elements was  at the top level to avoid their second evaluation in modifier.

src/sub-blocks/Content/Content.tsx
DC-RomanKarpov: The idea was to reuse this component in custom block. And some of the inner components should be rearranged (paddings should be changed a little). Maybe for that case it's better to split the Content components into a smaller pieces that render text/links/buttons.

PR #819, opened 2024-02-16
feat: add support `ContentList` in `BasicCard` and `ExtendedFeatures`

NikitaCG:


src/blocks/ExtendedFeatures/ExtendedFeatures.tsx
```diff
@@ -38,6 +38,7 @@ export const ExtendedFeaturesBlock = ({
                         ({
                             title: itemTitle,
                             text,
+                            list,
```
gorgeousvlad, line 41: how does it work with no changes in types?

NikitaCG: ’ExtendedFeaturesItem’ already use ’Omit<ContentBlockProps, >’

gorgeousvlad (approved):

PR #820, opened 2024-02-16
feat: migrate from doc tools to diplodoc transform

qradle-yndx:


package.json
```diff
@@ -165,6 +165,7 @@
     "jest-transform-css": "^6.0.0",
     "js-yaml": "^4.1.0",
     "lint-staged": "^11.2.6",
+    "markdown-it": "^14.0.0",
```
NikitaCG, line 168: you dont use this package

NikitaCG (changes requested):

package.json
qradle-yndx: @diplodoc/transform doesn't work without it. 
`ModuleNotFoundError: Module not found: Error: Can't resolve 'markdown-it/lib/token'`

qradle-yndx: fixed

NikitaCG (approved):

PR #862, opened 2024-03-14
feat: add withBorderOnScroll prop for Navigation component

goshander:


src/navigation/components/Navigation/Navigation.tsx
```diff
@@ -43,6 +49,8 @@ export const Navigation: React.FC<NavigationProps> = ({data, logo, className}) =
     const onSidebarOpenedChange = (isOpen: boolean) => setIsSidebarOpened(isOpen);
 
     useEffect(() => {
+        if (!withBorderOnScroll) return () => {};
```
NikitaCG, line 52: I think we are no need new props to solve your problem from issue

src/navigation/components/Navigation/Navigation.tsx
```diff
@@ -43,6 +49,8 @@ export const Navigation: React.FC<NavigationProps> = ({data, logo, className}) =
...
+
         const showBorderOnScroll = () => {
             if (!showBorder) {
```
NikitaCG, line 55: Change `!showBorder -> !withBorder` in this place, and if `window.scrollY === 0`, the border will disappear.
it is what you want ?)

NikitaCG (changes requested):

src/navigation/components/Navigation/Navigation.tsx
goshander: This PR is not related to the issue. Its two independents problems

Border no needed at all, like on this landing (css trick now)
https://ytsaurus.tech/

NikitaCG: ok

NikitaCG (approved):

PR #870, opened 2024-03-20
fix: error when using uikit in server code

smsochneg:
When importing server code in NodeJs environment some files (e.g. .css) are getting required in js files.

### DIFF
--- a/src/text-transform/utils.ts
+++ b/src/text-transform/utils.ts
@@ -1,7 +1,8 @@
 import transformYFM, {Options, Output} from '@diplodoc/transform';
-import {Lang} from '@gravity-ui/uikit';
 import sanitize from 'sanitize-html';
 import Typograf from 'typograf';
+
+import {Lang} from './types';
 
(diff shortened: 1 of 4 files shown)


d3m1d0v (dismissed):

src/text-transform/utils.ts
```diff
@@ -54,10 +55,10 @@ function enableRules(tp: typograf.Typograf) {
     disabled.forEach((rule) => tp.disableRule(rule));
 }
 
-export function typograf(text: string, lang: `${Lang}` = Lang.Ru) {
+export function typograf(text: string, lang: Lang = 'ru') {
     const localeByLang = {
-        [Lang.Ru]: ['ru', 'en-US'],
-        [Lang.En]: ['en-US', 'ru'],
+        ['ru']: ['ru', 'en-US'],
```
NikitaCG, line 60: now `[]` is  not necessary for key, use `{ru: [], en: []}`

NikitaCG (approved):

NikitaCG (approved):

PR #887, opened 2024-04-01
feat: allow to disable image slider inside media component

benax-se:


src/components/Media/Image/Image.tsx
```diff
@@ -34,16 +34,17 @@ export const defaultAnimatedDivQa = 'animated-div';
...
     const {
         parallax,
-        image,
         height,
         imageClassName,
         isBackground,
         hasVideoFallback,
         video,
         fullscreen,
+        disableImageSlider,
```
NikitaCG, line 43: maybe enhanced name -> disableImageSliderForArrayInput ?

NikitaCG (dismissed):

src/components/Media/Image/Image.tsx
benax-se: I had such an idea, but the slider appears only when we pass an array to image prop, so it seems that such a name is too verbose

NikitaCG: You know this fact because you've seen the code, someone who hasn't seen it might think that it's always a slider being drawn and passing this property even when not an array is passed

benax-se: you're right, i renamed it

NikitaCG (approved):

PR #889, opened 2024-04-01
feat: add media images fullscreen slider

qradle-yndx:


src/components/Media/Media.tsx
```diff
@@ -32,6 +33,7 @@ export const Media = (props: MediaAllProps) => {
         previewImg,
         parallax = false,
         fullscreen,
+        fullscreenImageCover,
```
NikitaCG, line 36: If this is a boolean, we should name it "hasFullscreenCover" or something similar

NikitaCG (changes requested):

src/components/Media/Media.tsx
qradle-yndx: I think `isFullscreenImageCover` would be more appropriate than `has` prefix, isn't it?

NikitaCG (approved):

PR #918, opened 2024-04-10
feat: update analyticsEvents handler conditioin: run on form redirect

niktverd:


src/components/YandexForm/YandexForm.tsx
```diff
@@ -112,7 +112,7 @@ const YandexForm = (props: YandexFormProps) => {
                     onLoad?.();
                 }
 
-                if (message === 'sent') {
```
NikitaCG, line 115: please add comment why we need to check redirectUrl

NikitaCG (approved):

src/components/YandexForm/YandexForm.tsx
niktverd: done

NikitaCG (approved):

PR #919, opened 2024-04-10
feat(ImageCard): add contentPosition

DC-RomanKarpov:
+ generalize contentPosition behavior for all cards

### DIFF
--- a/src/sub-blocks/LayoutItem/LayoutItem.tsx
+++ b/src/sub-blocks/LayoutItem/LayoutItem.tsx
@@ -2,7 +2,7 @@ import React, {useMemo} from 'react';
 
 import {useUniqId} from '@gravity-ui/uikit';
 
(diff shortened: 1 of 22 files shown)


src/sub-blocks/LayoutItem/LayoutItem.tsx
```diff
@@ -15,7 +15,7 @@ import './LayoutItem.scss';
 const b = block('layout-item');
 
 const LayoutItem = ({
-    content: {links, buttons, ...content},
```
NikitaCG, line 18: why are we removing the buttons?

NikitaCG (changes requested):

src/sub-blocks/LayoutItem/LayoutItem.tsx
DC-RomanKarpov: Because we pass them to the `Content` component directly. We do not make any transformations with them like with links.

NikitaCG (approved):

PR #924, opened 2024-04-11
feat: add media images fullscreen slider

qradle-yndx:


src/components/FullscreenImage/FullscreenImage.tsx
```diff
@@ -15,19 +18,38 @@ export interface FullscreenImageProps extends ImageProps {
...
+        extraProps,
+    } = props;
     const [isOpened, setIsOpened] = useState(false);
+    const [sliderLoaded, setSliderLoaded] = useState(false);
 
     const openModal = () => setIsOpened(true);
     const closeModal = () => setIsOpened(false);
 
+    useEffect(() => {
+        if (!isOpened) {
```
NikitaCG, line 44: add condition if we have sliderData, because we dont need trigger this hook without slider

NikitaCG (changes requested):

PR #939, opened 2024-04-27
fix: add new option in content transform callback

NikitaCG:


src/components/Image/schema.ts
```diff
@@ -76,6 +76,20 @@ export const ImageProps = {
             ...ImageDeviceProps,
             optionName: 'device options',
         }),
+        {
```
NikitaCG, line 79: for these cases 
```
background:
  image:
    - when: <condition>
      desktop: <desktop_url>
      mobile: <mobile_url>
    - when: <condition>
          desktop: <desktop_url>
          mobile: <mobile_url>
      ```

gorgeousvlad (approved):

PR #974, opened 2024-08-01
fix: feature icons a11y

PahaN47:


src/blocks/ExtendedFeatures/ExtendedFeatures.tsx
```diff
@@ -57,7 +57,9 @@ export const ExtendedFeaturesBlock = ({
 
                             return (
                                 <Col className={b('item')} key={text || itemTitle} sizes={colSizes}>
-                                    {iconData && <Image {...iconData} className={b('icon')} />}
+                                    <div className={b('icon-wrap')} aria-hidden>
```
NikitaCG, line 60: If we dont have `iconData`, we are no needed wrapper

NikitaCG (changes requested):

NikitaCG (approved):

PR #980, opened 2024-08-07
fix: filter block a11y

PahaN47:


src/sub-blocks/Content/Content.tsx
```diff
@@ -67,10 +68,11 @@ const Content = (props: ContentProps) => {
                     title={titleProps}
                     colSizes={{all: 12}}
                     id={titleId}
+                    {...a11yProps?.title}
```
NikitaCG, line 71: why we need `a11yProps` if we have static role for element ?title always title, list always list and e.t.c

NikitaCG (changes requested):

src/sub-blocks/BasicCard/BasicCard.tsx
```diff
@@ -37,7 +33,7 @@ const BasicCard = (props: BasicCardProps) => {
             className={b()}
             contentClassName={b('content')}
             {...cardParams}
-            extraProps={{'aria-describedby': descriptionId, 'aria-labelledby': titleId}}
+            extraProps={{role: 'group'}}
```
Kyzyl-ool, line 36: Why the card (especially [card with url](https://preview.gravity-ui.com/page-constructor/980/?path=/story/components-cards-basiccard--with-url)) should have `group` role? If one needs the `group` role specified, then it should be possible to specify it via props, e.g. via `extraProps`.

src/sub-blocks/BasicCard/BasicCard.tsx
```diff
@@ -46,16 +42,17 @@ const BasicCard = (props: BasicCardProps) => {
...
-                        textId={descriptionId}
                         additionalInfo={additionalInfo}
                         links={links}
                         list={list}
                         buttons={buttons}
                         size="s"
                         colSizes={{all: 12, md: 12}}
                         controlPosition={areControlsInFooter ? 'bottom' : 'default'}
+                        a11yProps={{
+                            title: {role: 'link', tabIndex: 0},
```
Kyzyl-ool, line 54: If you want `link` role, it is better to use `<a>` tag: no need to specify the `role` and `tabIndex`. But be aware of layout change due to browser default styles

Kyzyl-ool (changes requested):

src/sub-blocks/Content/Content.tsx
PahaN47: In this specific case its here to give `Title` `role="link"`

PahaN47:
To address the comments:
The expected way of interacting with cards is not interacting with the card itself, but interacting with its contents sequentially.
In testing I noticed that the user would focus on the entire card, then they would hear the title and go on to the next card. It appears that descriptions are not read out loud by default in some screen reader programs.
When the user was told there was a description that they missed only then they were able to find and access it.
Then the user knowing the way cards are usually handled assumed that they were focusing on a card title which was a link and it had `aria-describedby` with the card description. So they commented that there is no need to add the description to title's `aria-desccribedby` as it is normal to traverse cards by focusing on both their title link and description one after another.
What the user had no way of knowing is that they had never focused on neither the card's title nor it's description. They were focused on the entire card element with `aria-label` containing title.
The goal of these changes was to bring the link-wrapped card up to the expected accessibility standard: a group of elements where title is a link and description is the following text that can be focused by screen readers as a separate element. Generally wrapping a card inside an `<a>` tag is considered a bad a11y practice, but changing that inside the `CardBase` component would be a potentially breaking change.
There may be flaws in my implementation, but the expected result nonetheless should be:
- card itself is not focusable
- title is a link
- description is text that follows the title and is a separate element

PahaN47:
Removed changes related to cards as they are a separate issue now #998

NikitaCG (approved):

Kyzyl-ool (approved):

PR #982, opened 2024-08-08
fix: slider a11y

PahaN47:


src/blocks/Slider/Slider.tsx
```diff
@@ -8,6 +8,7 @@ import React, {
     useState,
```
NikitaCG, line 8: In new slider we have same problems ? you need to check it and fix if we need

NikitaCG (changes requested):

src/blocks/Slider/Slider.tsx
```diff
@@ -289,14 +307,20 @@ export const SliderBlock = (props: React.PropsWithChildren<SliderProps>) => {
...
+        const isVisible = isVisibleSlide(index);
         return (
             <li
                 key={index}
                 className={b('dot', {active: index === currentIndex})}
                 onClick={() => handleDotClick(index)}
-                aria-hidden={isVisibleSlide(index) ? true : undefined}
-                aria-label={`Slide ${getCurrentSlideNumber(index)} of ${barSlidesCount}`}
-            ></li>
+                role="button"
```
Kyzyl-ool, line 316: `li` should have only `listitem` or `generic` role. It is not recommended to change it: https://www.w3.org/TR/html-aria/#el-li

src/blocks/Slider/Slider.tsx
```diff
@@ -289,14 +307,20 @@ export const SliderBlock = (props: React.PropsWithChildren<SliderProps>) => {
...
+                aria-hidden={isVisible ? true : undefined}
+                tabIndex={isVisible ? -1 : 0}
```
Kyzyl-ool, line 318: It is preferred to implement the so called "Roving tabindex" https://www.tohuwabohu.io/2022/10/roving-tabindex/#menus-and-interactive-lists

It is much better, as we can focus only on 1 dot instead of all visible ones and change the active tab using arrow keys

PahaN47:
There was a problem with current changes causing autofocus with autoplay. The handling of autoplay is different between the two versions now (partially because of some things not working out on both sliders). The old slider differentiates causes focus on user input only.
However, I thought that autoplay as a concept is very confusing for a visually impaired person, so with the new slider when autoplay is on there is no autofocus at all, the previous/next/pagination buttons are hidden and the user can select any item whether is is actually visible or not as if it was just a list of items

src/blocks/SliderNew/utils.ts
```diff
@@ -47,3 +52,198 @@ export function getSliderResponsiveParams({
...
+    useEffect(() => {
+        setMemoizedValue((memoized) =>
+            typeof value in MEMOIZABLE_TYPES && isEqual(memoized, value) ? memoized : value,
+        );
+    }, [value]);
+
+    return memoizedValue;
+};
+
+interface PaginationBulletData {
```
qradle-yndx, line 71: Is it possible/useful to use `bullets` prop from `{PaginationMethods} from 'swiper/types'`?

PahaN47: It is useful to access the DOM elements, though it is typed wrongly as Dom7Array[]. It is used inside usePaginationBullets() to form PaginationBulletData. This interface keeps the bullets together with their event handlers for the latter to be removed after rerenders

qradle-yndx (dismissed):

Kyzyl-ool:
As I can see, in the new slider there each card has 2 focusable tags: <a> and <div> with tabindex=0. Maybe one should not set tabindex=0 in case if the card is the link?

https://preview.gravity-ui.com/page-constructor/982/?path=/story/lab-slidernew--default

Kyzyl-ool:
And when I navigate through the cards with tabbing the dots panel disappears(

Kyzyl-ool:
https://preview.gravity-ui.com/page-constructor/982/?path=/story/blocks-slider--default

When one presses on the Left/Right buttons while focused on the dot, the slides should change to prev/next ones accordingly

PahaN47:
> When one presses on the Left/Right buttons while focused on the dot, the slides should change to prev/next ones accordingly, not just switch the focus to the prev/next dot

This contradicts the behavior I was going for initially, which is to focus on the first new slide once the page is changed. This was proposed by one of our a11y testers

PahaN47:
> And when I navigate through the cards with tabbing the dots panel disappears(

Inside the new slider this has been fixed by giving invisible slides `visibility: hidden` so than nothing inside can be tabbed to. There does not seem to be a way to fix the moving dots problem itself as far as I can see, so with autoplay on where any slide can be tabbed to it is still present.

Kyzyl-ool (approved):

PahaN47:
After discussion with @NikitaCG it was decided that implementation of roving tabindex is adding too much complexity to an otherwise simple component. So I've tried to remove it and simplify everything else inside the new slider.

src/blocks/SliderNew/useSlider.tsx
```diff
@@ -82,3 +85,39 @@ export const useSlider = ({
         },
     };
 };
+
+export const useSliderPagination = (props: {
```
NikitaCG, line 89: put in different file `useSliderPagination`

src/blocks/SliderNew/Slider.tsx
```diff
@@ -124,18 +153,22 @@ export const SliderNewBlock = ({
...
-                            onClick={onNext}
-                            size={arrowSize}
-                        />
+                        <div aria-hidden={withAutoplay}>
+                            <Arrow
+                                className={b('arrow', {prev: true})}
+                                type="left"
+                                onClick={onPrev}
+                                size={arrowSize}
+                                extraProps={{tabIndex: withAutoplay ? -1 : 0}}
```
NikitaCG, line 162: `{tabIndex: withAutoplay ? -1 : 0}`  save in constant  because use in different places

src/blocks/SliderNew/utils.ts
```diff
@@ -47,3 +50,72 @@ export function getSliderResponsiveParams({
...
+        case element instanceof HTMLTextAreaElement:
+        case element instanceof HTMLButtonElement:
+            htmlElement = element as HTMLSelectElement | HTMLTextAreaElement | HTMLButtonElement;
+            return !htmlElement.disabled;
+        default:
+            return false;
+    }
+};
+
+export const getFocusAnchor = (element: HTMLElement | null) => {
```
NikitaCG, line 106: delete, because not used

src/blocks/SliderNew/useSlider.tsx
```diff
@@ -1,23 +1,26 @@
...
+
+export const useSlider = ({children, autoplayMs, type, ...props}: UseSliderProps) => {
     const [slider, setSlider] = useState<Swiper>();
     const [isLocked, setIsLocked] = useState(false);
+    const slidesToShow = useMemoized(props.slidesToShow);
 
     const childrenCount = React.Children.count(children);
 
-    const autoplayEnabled = useMemo(() => Boolean(autoplayMs), [autoplayMs]);
+    const autoplayEnabled = autoplayMs !== undefined && autoplayMs > 0;
```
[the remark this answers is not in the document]
PahaN47: it would be enough if it wasn't for typescript

NikitaCG (changes requested):

NikitaCG (approved):

PR #984, opened 2024-08-12
feat: add customization to mobile navigation

JeikZim:
# Add customization to mobile navigation

## Summary

Several features have been added that together allow you to make a custom mobile menu, add elements to the header, and link your custom components with your own logic using states, etc.

## Feat 1. Custom header items on mobile 
Added the ability to insert items into the mobile header using NavigationList and OverflowScroller.

### Preview

**IPhone XR**
feat 1 on IPhone XR 1
feat 1 on IPhone XR 2
feat 1 on IPhone XR 3

https://github.com/user-attachments/assets/539e65fb-f7cd-496f-89c7-cdbe66960ee1

**IPad Mini**
feat 1 on IPad Mini

### How to add 

```
const mobileHeaderItems = [{type: 'item1'}, {type: 'item2'}, {type: 'item3'}];

<PageConstructor
    custom={{
        navigation: {
            item3: () => ...,
        },
    }
    ...
    navigation={
         header: {
    ...
                 // new
                 customMobileHeaderData: {
                         mobileHeaderItems,
                  },
    ...
             }
        }
/>
```

## Feat 2. CustomMobileMenuButton
Added the ability to use a custom mobile menu button.

demo of custom mobile menu button

### How to add 

```
const MobileMenuButton = () => <SomeComponent />

<PageConstructor
    ...
    navigation={
         header: {
    ...
                 // new
                 customMobileHeaderData: {
                         CustomMobileMenuButton: MobileMenuButton,
                  },
    ...
             }
        }
/>
```

## Feat 3. CustomMobileMenu

Added the ability to use a full-screen mobile menu, with custom Header, Content and Footer components, which are optional. It is also possible to choose the direction of the menu appearance.

**FullscreenMobileNavigation with Header (1), Content (2) and Footer (3)**
Снимок экрана 2024-08-12 в 12 43 39

Without CustomContent (use basic content)
Снимок экрана 2024-08-12 в 12 44 57

### How to add 

```
const MobileMenuButton = () => <SomeComponent />

<PageConstructor
    ...
    navigation={
         header: {
    ...
                 // new
                 customMobileMenuData: {
                         appearanceSide: 'left', 
                         Header: SomeHeaderComponent',
                         Content: SomeContentComponent',
                         Footer: SomeFooterComponent,
                  },
    ...
             }
        }
/>
```

## Feat 4

Added the ability to transfer a custom hook to navigation. The result of executing this hook is transmitted strictly to custom components (from this PR), and does not affect anything else.

This feature allows you to create interaction between components (for example, the Header from the mobile menu and CustomMobileMenuButton), which will occur at the level of the Navigation component. This will add more interactivity to the menu on mobile devices.

### How to add 

```
const MobileMenuButton = () => <SomeComponent />

<PageConstructor
    ...
    navigation={
         header: {
    ...
                 // new
                 useCustomHook: useSomeCustomHook,
    ...
             }
        }
/>
```

### DIFF
--- a/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
+++ b/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
@@ -12,7 +12,7 @@ $block: '.#{$ns}fullscreen-mobile-navigation';
     display: flex;
     flex-direction: column;
     width: 100%;
-    height: calc(100vh - var(--header-height) + 1px);
+    height: calc(100vh - var(--header-height) + 1);
 
     background-color: var(--g-color-base-background);
     box-shadow: 0 3px 10px var(--g-color-sfx-shadow);
--- a/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
+++ b/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
@@ -12,7 +12,7 @@ $block: '.#{$ns}fullscreen-mobile-navigation';
     display: flex;
     flex-direction: column;
     width: 100%;
-    height: calc(100vh - var(--header-height));
+    height: calc(100vh - var(--header-height) + 1px);
 
     background-color: var(--g-color-base-background);
     box-shadow: 0 3px 10px var(--g-color-sfx-shadow);
--- a/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
+++ b/src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
@@ -12,7 +12,7 @@ $block: '.#{$ns}fullscreen-mobile-navigation';
     display: flex;
     flex-direction: column;
     width: 100%;
-    height: calc(100vh - var(--header-height) + 1);
+    height: calc(100vh - var(--header-height) + 1px);
 
     background-color: var(--g-color-base-background);
     box-shadow: 0 3px 10px var(--g-color-sfx-shadow);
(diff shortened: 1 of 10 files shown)


src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    // * disable page scroll
+    body:has(&_opened) {
+        overflow-y: hidden;
+    }
+
+    // * appear directions
+    &_left {
+        left: -105vw;
```
martyanovandrey, line 99: Can you please write in more detail why we use there are variables with viewport

src/navigation/components/Navigation/Navigation.tsx
```diff
@@ -26,7 +27,15 @@ export const Navigation: React.FC<NavigationProps> = ({data, logo, className}) =
         iconSize = 20,
         withBorder = false,
         withBorderOnScroll = true,
+        useCustomHook = () => {
```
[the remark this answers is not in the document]
JeikZim: In this case, I need to get an empty object if the use Custom Hook function was not passed. Therefore, the default value is a function that returns an empty object.

It `useCustomHook = () => {}` differs from this `useCustomHook = () => { return {} }` in that in the first case an `undefined` is returned, and in the second case, an empty object `{}` is returned.

But to make the code cleaner, I changed the code here:
`useCustomHook = () => { return {} } ` on `useCustomHook = () => {} `

and here:
`const customHookData = useCustomHook();` on `const customHookData = useCustomHook() || {};`

So it is clearly seen that if the hook was not passed, an empty object will be used.

src/navigation/components/FullscreenMobileNavigation/FullscreenMobileNavigation.scss
JeikZim: We decided to keep this logic, but rename the name of the prop

PR #988, opened 2024-08-16
fix(Content): apply control-position: bottom to links

ariser:
classname being wrong prevented the proper alignment of links in Content
`controlPosition: bottom` doesn't work for them

I can see that in the same file there are other instances of `#{$block}__link`. This classname doesn't exist in the component, so they do nothing, and either cause bugs, or should be removed entirely. Depending on the intention. I don't know how it should be, so only fixed the one I understand.


ariser:
Roma is having a day off today :(

DC-RomanKarpov (approved):

ariser:
@NikitaCG can you approve this?
it's a single character PR
Roma's approve doesn't do it

NikitaCG (approved):

PR #990, opened 2024-08-19
refactor: replace <button> to <Button> from uikit

JeikZim:


NikitaCG:
@JeikZim  please rename this PR like `feat:`

NikitaCG (changes requested):
https://github.com/gravity-ui/uikit/blob/main/src/components/Button/Button.scss
Control sizes are  not equal uikit Button sizes

JeikZim:
> https://github.com/gravity-ui/uikit/blob/main/src/components/Button/Button.scss Control sizes are not equal uikit Button sizes

It really is. However, is this a problem? This component is used only in one place (in MobileMenuButton), and there it has an "l" value equal to this value in uikit

NikitaCG:
> > https://github.com/gravity-ui/uikit/blob/main/src/components/Button/Button.scss Control sizes are not equal uikit Button sizes
> 
> It really is. However, is this a problem? This component is used only in one place (in MobileMenuButton), and there it has an "l" value equal to this value in uikit

Yes, it's problem because it's breaking change for our users

We are need to release it in major version and receive 'ok' from design

PR #993, opened 2024-08-27
feat!: yfm instead of html

niktverd:


src/sub-blocks/PriceCard/PriceCard.scss
```diff
@@ -63,8 +63,10 @@ $block: '.#{$ns}price-card';
     }
 
     &__title {
-        @include text-size(subheader-3);
-        margin-bottom: $indentSM;
+        @include add-specificity(&) {
```
NikitaCG, line 66: I see more `specificity` in this pr why they is appear ?)

src/sub-blocks/BannerCard/BannerCard.scss
```diff
@@ -57,12 +59,14 @@ $block: '.#{$ns}banner-card';
...
 
     &__subtitle {
-        display: inline-block;
-        margin-top: 8px;
-        @include subtitle();
-
-        & p {
+        @include add-specificity(&) {
+            display: inline-block;
+            margin-top: 8px;
```
NikitaCG, line 64: I think we can refactor this place on Indent variable

src/blocks/Questions/QuestionBlockItem/QuestionBlockItem.scss
```diff
@@ -12,19 +12,25 @@ $block: '.#{$ns}QuestionsBlockItem';
...
+            a {
+                @include link();
+            }
         }
     }
 
+    &__title-container {
+        margin: 0;
+        position: relative;
+        padding-right: 24px;
```
NikitaCG, line 31: `$indent`

NikitaCG (changes requested):

src/sub-blocks/PriceCard/PriceCard.scss
niktverd: Because, React.createElement works with classes slightly different. And styles of yfm override styles from component classes

src/components/HTML/HTML.tsx
```diff
@@ -1,35 +1,61 @@
...
 
-import {QAProps} from '../../models/common';
-import {hasBlockTag} from '../../utils';
+import {ClassNameProps, QAProps} from '../../models/common';
+import {selectVariant} from '../../utils';
 
-export interface HTMLProps {
-    children?: string;
+export interface HTMLExtraProps {
+    variant?: 'span' | 'div' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'section' | 'p';
```
NikitaCG, line 7: would it be easier to make an enum for variand use it in other places like this https://github.com/gravity-ui/page-constructor/pull/993/files#diff-8a573455e8636f1d6dd641125e9b47de96bbf75f3fdd9bff39a2233e7515144bR27 ?

NikitaCG, line 7: and I think we need more informative name for this props , maybe `tagVariant` or `tagType`

NikitaCG (changes requested):

src/components/HTML/HTML.tsx
niktverd: changed it to tagName

niktverd: enum is good if value is unknown, or difficult to write. When value is simple and well-known it is better to use simple type as I do.

src/blocks/Questions/QuestionBlockItem/QuestionBlockItem.scss
```diff
@@ -12,14 +12,26 @@ $block: '.#{$ns}QuestionsBlockItem';
     }
 
     &__title {
-        @include reset-button-style();
-        @include heading4();
-        @include focusable();
+        @include add-specificity(&) {
+            // @include reset-button-style();
```
NikitaCG, line 16: delete if we dont need it

NikitaCG (approved):

NikitaCG (approved):

PR #995, opened 2024-08-29
fix: image card link hover

qradle-yndx:


NikitaCG (approved):
rename to feat please

Issue #961, opened 2024-07-22
HeaderBreadcrumbs: possible double space around the separator


Lunatic174:
In the HeaderBreadcrumbs component, there might be a double space around the separator in the breadcrumbs.

<img width="459" alt="image" src="https://github.com/user-attachments/assets/62f72e8d-8be3-4975-8389-c3d7b0d5fef1">

The separator has the following [margin](https://github.com/gravity-ui/page-constructor/blob/0cb9594fe6398d397845cb46e503e80656484103/src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.scss#L21) settings: `margin: 0 8px 0 6px;`. 
Additionally, there is a space in the [content](https://github.com/gravity-ui/page-constructor/blob/0cb9594fe6398d397845cb46e503e80656484103/src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.scss#L20C13-L20C30) of the pseudo-element: `content: '\a0/ ';` with spaces before and after the `/` character. 
This results in a double spacing. 
This might be an error, and one of these should be removed to fix the issue.

PR #996, opened 2024-08-30
fix: breadcrumbs a11y

PahaN47:


src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.scss
```diff
@@ -6,27 +6,29 @@ $block: '.#{$ns}header-breadcrumbs';
...
-
         &:hover {
             color: var(--g-color-text-primary);
         }
     }
 
+    &__separator {
+        margin: 0 8px 0 6px;
+        color: var(--g-color-text-secondary);
+    }
```
NikitaCG, line 30: lost color for dark theme https://preview.gravity-ui.com/page-constructor/996/?path=/story/blocks-header--breadcrumbs

see in main branch version https://preview.gravity-ui.com/page-constructor/?path=/story/blocks-header--breadcrumbs

src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.tsx
```diff
@@ -11,22 +11,34 @@ import './HeaderBreadcrumbs.scss';
 const b = block('header-breadcrumbs');
 
 export default function HeaderBreadcrumbs(props: HeaderBreadCrumbsProps) {
-    const {items, theme = 'light', className, analyticsEvents} = props;
+    const {items, theme = 'light', className, analyticsEvents, currentPageUrl} = props;
```
NikitaCG, line 14: As I understand it, this prop will need to be manually registered in the page configuration, I'm not sure that someone will do this every time in order to get aria-current. We need some kind of solution out of the box so that the component understands what the URL is now and does your check. Also, we must not forget that our components are used in ssr

NikitaCG (changes requested):

src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.scss
NikitaCG: not fixed

src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.tsx
```diff
@@ -10,28 +12,52 @@ import './HeaderBreadcrumbs.scss';
...
+const isCurrentPage = (url: string): boolean => {
+    if (url === '' || url === '.' || url.startsWith('#')) {
+        return true;
+    }
+
+    if (typeof window === 'undefined' || url.startsWith('.')) {
+        return false;
+    }
+
+    const location = window.location;
```
NikitaCG, line 24: use window only in useEffect please

In hook you need to set current page url state and next, in your function you need to compare current url with item.url

src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.tsx
```diff
@@ -10,28 +12,52 @@ import './HeaderBreadcrumbs.scss';
...
+    }
+
+    return false;
+};
+
 export default function HeaderBreadcrumbs(props: HeaderBreadCrumbsProps) {
-    const {items, theme = 'light', className, analyticsEvents, currentPageUrl} = props;
+    const {items, theme = 'light', className, analyticsEvents} = props;
     const handleAnalytics = useAnalytics(DefaultEventNames.Breadcrumb);
+    const forceUpdate = useForceUpdate();
```
NikitaCG, line 40: we are no need to use this

NikitaCG (changes requested):

src/components/HeaderBreadcrumbs/HeaderBreadcrumbs.tsx
PahaN47: we are in need of doing it this way specifically, because i cannot think of a simple and guaranteed way of subscribing to `window.location` changes. thus if there is a transition between pages in an SPA, we have no way of updating our current location with it.

PahaN47:
Fixing this as well https://github.com/gravity-ui/page-constructor/issues/961#issue-2422777560

NikitaCG (approved):

PR #999, opened 2024-09-04
feat: add videoIframe prop to media block and component

benax-se:


src/components/VideoBlock/VideoBlock.tsx
```diff
@@ -84,25 +89,29 @@ const VideoBlock = (props: VideoBlockProps) => {
     } = props;
     const handleAnalytics = useAnalytics(DefaultEventNames.VideoPreview);
 
-    const src = getVideoSrc(stream, record);
+    const src = videoIframe ? videoIframe : getYoutubeVideoSrc(stream, record);
     const ref = useRef<HTMLDivElement>(null);
-    const iframeRef = useRef<HTMLIFrameElement>();
     const [hidePreview, setHidePreview] = useState(false);
-    const norender = (!stream && !record) || !src;
+    const norender = !src;
```
NikitaCG, line 95: I think we are no need to set this value

NikitaCG (changes requested):

src/components/VideoBlock/VideoBlock.tsx
benax-se: refactored

NikitaCG (approved):

PR #1014, opened 2024-09-12
feat: visual tests for blocks with yfm

niktverd:


NikitaCG:
I think this is not feat, please change  pr name to test or chore

PR #1025, opened 2024-09-24
feat: add className prop for banner block

aleksey-shkadov:


src/blocks/Banner/Banner.tsx
```diff
@@ -10,10 +10,12 @@ import './Banner.scss';
...
 
 export const BannerBlock = (props: BannerBlockProps) => {
-    const {animated, ...bannerProps} = props;
+    const {animated, className, ...bannerProps} = props;
+
+    const divClassName = b(null, className);
 
     return (
-        <AnimateBlock className={b()} animate={animated}>
+        <AnimateBlock className={divClassName} animate={animated}>
```
NikitaCG, line 18: use immediately  `className={b(null, className)}`

NikitaCG (changes requested):

src/blocks/Banner/Banner.tsx
aleksey-shkadov: Done

PR #1057, opened 2024-10-24
fix(ReactPlayer): change default contain behaviour

PahaN47:


src/components/ReactPlayer/ReactPlayer.tsx
```diff
@@ -89,7 +89,7 @@ export const ReactPlayerBlock = React.forwardRef<ReactPlayerBlockHandler, ReactP
             ariaLabel,
             ratio,
             autoRatio,
-            contain,
+            contain = true,
```
aeksandla, line 92: This looks like a non-backward compatible change.

NikitaCG: Why is this change necessary?

aeksandla (approved):

PR #1063, opened 2024-11-05
fix(QuestionBlock): duplicate faq schema

PahaN47:


src/blocks/Questions/models.ts
```diff
@@ -1,5 +1,6 @@
 export const FaqMicrodataValues = {
     PageType: 'https://schema.org/FAQPage',
+    PageId: '/faq',
```
NikitaCG, line 3: if we have multiple FAQ's, will everyone have the same one? Is this normal for robots?

NikitaCG (changes requested):

src/blocks/Questions/models.ts
PahaN47: the point was to give different faq's links to the same page. google's rich results test does not seem to have a problem with it

NikitaCG (approved):

Issue #966, opened 2024-07-29
Fix `jsx-a11y/no-noninteractive-element-to-interactive-role` rule


NikitaCG:
- [ ] https://github.com/gravity-ui/page-constructor/blob/main/src/blocks/Questions/QuestionBlockItem/QuestionBlockItem.tsx#L38

Issue #964, opened 2024-07-29
Fix `jsx-a11y/alt-text` rule


NikitaCG:
- [ ] https://github.com/gravity-ui/page-constructor/blob/main/src/components/ImageBase/ImageBase.tsx#L0-L1
- [ ] https://github.com/gravity-ui/page-constructor/blob/main/src/sub-blocks/PriceDetailed/PriceDetails/Details/List.tsx

PR #1069, opened 2024-11-08
fix: eslint a11y rules

PahaN47:


PahaN47:
trying to fix these issues: #964 #965 #966

PR #1072, opened 2024-11-11
feat: add server schema build & export

martyanovandrey:


schema.webpack.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-env node */
+const path = require('path');
```
NikitaCG, line 2: Please provide a comment with information on why we need this build.

NikitaCG (dismissed):

NikitaCG (approved):

PR #1084, opened 2024-11-27
feat: brand footer

PahaN47:


aeksandla (dismissed):

src/containers/PageConstructor/PageConstructor.tsx
```diff
@@ -50,6 +51,7 @@ export interface PageConstructorProps {
     custom?: CustomConfig;
     renderMenu?: () => React.ReactNode;
     navigation?: NavigationData;
+    branded?: boolean;
```
NikitaCG, line 54: `isBranded`
and add info in readme https://github.com/gravity-ui/page-constructor/blob/main/README.md#L42

NikitaCG (changes requested):

aeksandla (approved):

src/icons/BrandIcon.tsx
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
+
+import {useTheme} from '../context/theme';
+import {Theme} from '../models';
+import {a11yHiddenSvgProps} from '../utils/svg';
+
+export const BrandIcon: React.FC<React.SVGProps<SVGSVGElement>> = (props) => {
+    const theme = useTheme();
```
NikitaCG, line 8: it seems to me that we need to create two icons and decide which one to use in this [place](https://github.com/gravity-ui/page-constructor/pull/1084/files#diff-5af56b1085dd18b3dc134f9c9bd4e4a927a53823224377619050802b802d0828R21). I think it will be more obvious why it is changing.

NikitaCG (changes requested):

NikitaCG (approved):

PR #1132, opened 2025-02-25
feat(mobileNavigation): added a parameter for the portal in mobile navigation

ReFFaT:


src/navigation/components/Navigation/Navigation.tsx
```diff
@@ -25,6 +25,7 @@ export const Navigation: React.FC<NavigationComponentProps> = ({data, logo, clas
         iconSize = 20,
         withBorder = false,
         withBorderOnScroll = true,
+        mobilePortalContainer,
```
NikitaCG, line 28: I don't think this prop refers to data that can be edited in the config. I think it should be placed on this [level](https://github.com/gravity-ui/page-constructor/pull/1132/files#diff-1d2e46086bba5a81bc57ae2c61307989c649b976c39c0f22eec936554516df5bR17)

NikitaCG (changes requested):

NikitaCG (approved):

PR #1157, opened 2025-04-08
feat(FoldableList): add new block

aeksandla:


aeksandla:
@NikitaCG 
Maybe we should show this block in the Lab section in storybook, since we will definitely modify it?

NikitaCG (approved):

PR #1165, opened 2025-04-10
fix(Slider): fix react-slick import

kseniya57:
Screenshot 2025-04-10 at 16 17 36


PahaN47 (changes requested):
As far as I understand, you did not experience the same import issues with `react-player`? If so, this means the problem may vary depending on build configuration and I suggest the fix should be applied to both modules anyway.

I checked the suggested fix on my build and it did not help with the problem. Perhaps something more universal would help. For example:
```
import _SlickSlider from 'react-slick';
const SlickSlider = _SlickSlider.default ?? _SlickSlider;
```

PahaN47 (approved):

PR #1182, opened 2025-04-24
feat(Image): add hide param

aeksandla:


src/components/Image/Image.tsx
```diff
@@ -61,48 +80,64 @@ const Image = (props: ImageProps) => {
...
+        !src ||
         projectSettings.disableCompress ||
         disableCompress ||
         !isCompressible(src) ||
         imgLoadingError;
 
     return (
         <picture className={containerClassName} data-qa={qa}>
-            {mobile && (
+            {(mobile || hideDevices.mobile) && (
```
NikitaCG, line 116: if `hideDevices.mobile === true` I think we don't need render this block,  but It's the opposite now
maybe we need to change name for `hideDevices`

src/components/Image/Image.tsx
```diff
@@ -61,48 +80,64 @@ const Image = (props: ImageProps) => {
...
-                    disableWebp={disableWebp}
-                    breakpoint={BREAKPOINTS.md}
+                    src={tablet || EMPTY_IMG}
+                    disableWebp={disableWebp || Boolean(hideDevices.tablet)}
+                    maxBreakpoint={BREAKPOINTS.md}
+                    minBreakpoint={BREAKPOINTS.sm}
                     qa={qaAttributes.tabletSource}
                 />
             )}
+            {hideDevices.desktop && (
```
NikitaCG, line 133: where case for `hideDevices` is boolean `true` ?

NikitaCG (changes requested):

src/components/Image/Image.tsx
aeksandla: In line 135 it says that an empty picture is displayed here. If the image on the desktop is hidden, then we show an empty picture.

src/components/Image/Image.tsx
aeksandla: We show this block when the picture in mobile is hidden to show an empty picture. If you do not show this block when the `hideDevices.mobile === true` parameter, a desktop or tablet image will be installed on the mobile resolution.

PR #1183, opened 2025-04-25
chore: add memory bank for cline

NikitaCG:


NikitaCG:
I made a memory bank, I also made dependency graphs for several components and blocks, work will continue in this direction, I made sure that someone else besides me could do this.

PR #1188, opened 2025-05-13
fix: hydration tab caption

NikitaCG:


PahaN47 (approved):
Really?

NikitaCG:
> Really?

 HTML component generate another <p>. In result we have p in p and hydration error

aeksandla (approved):

PR #1197, opened 2025-05-29
refactor: Props with children fix (After removing React.FC)

AadeshMndr:
Removed `React.PropsWithChildren` from those components where `children` was not relavant.


AadeshMndr:
I've made the changes you requested !

I've removed `React.PropsWithChildren` from the components that did not use `children`

PR #1198, opened 2025-06-06
feat(NavigationLink): add handleAnalytics

Valent1325:


aeksandla (approved):

src/navigation/components/NavigationItem/components/NavigationLink/NavigationLink.tsx
```diff
@@ -17,9 +18,26 @@ type NavigationLinkProps = NavigationItemProps & NavigationLinkItem;
...
+        analyticsEvents,
+        ...rest
+    } = props;
+
     const linkExtraProps = getLinkProps(url, hostname, target);
     const iconData = icon && getMediaImage(icon);
+    const handleAnalytics = useAnalytics(DefaultEventNames.Link, url);
+
+    const onClick = () => {
+        handleAnalytics(analyticsEvents);
```
niktverd, line 39: Should e.preventDefault() be here?

Valent1325: No, e.preventDefault() should not be added. The onClick handler is only used for analytics tracking, and adding e.preventDefault() would break the navigation functionality of the links. The current implementation correctly allows both analytics tracking and navigation to occur as intended.

niktverd (approved):

PR #1228, opened 2025-07-03
feat(Header, ContentLayout): calculate isMobile by screen size

PahaN47:


src/context/windowWidthContext/BreakpointContext.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
```
aeksandla, line 1: why don't we use the useWindowBreakpoint hook?

src/blocks/HeaderSlider/HeaderSlider.tsx
```diff
@@ -12,7 +12,7 @@ const b = block('header-slider-block');
 
 export const HeaderSliderBlock = ({items, arrows, ...props}: HeaderSliderBlockProps) => {
     const isMobile = React.useContext(MobileContext);
-    const showArrows = isMobile ? false : arrows;
+    const showArrows = !isMobile && arrows;
```
NikitaCG, line 15: why ?

NikitaCG (changes requested):

src/context/windowWidthContext/BreakpointContext.tsx
PahaN47: It returns the first breakpoint that the screen width is larger or equal to. As I have looked at the styles, a have noticed that we **generally** tend to use something like:
```
@media (max-width: map-get($gridBreakpoints, 'sm'))
```
Which corresponds to `windowWidth <= BREAKPOINTS.sm`. This is not possible to do using  `useWindowBreakpoint` without modifying it in some way. Furthermore, this hook creates a separate `resize` listener for each use, which seems redundant.
Thus, rather than creating an opposite of `useWindowBreakpoint` or modifying it I decided to create a more practical and flexible solution for internal use.
Overall I think that `useWindowBreakpoint` must be reworked in a new major version and some styles like
```
@media (max-width: map-get($gridBreakpoints, 'sm') - 1)
```
need to be changed to achieve consistency with the rest.

src/blocks/HeaderSlider/HeaderSlider.tsx
PahaN47: Why not though?

NikitaCG: First variant is simple for reading in my opinion. But main point is what nothing change

src/context/windowWidthContext/BreakpointContext.tsx
NikitaCG: ```Overall I think that useWindowBreakpoint must be reworked in a new major version and some styles like```
create issue for this and we discuss it

Issue #1286, opened 2025-09-14
Video in media block has no loop


Marginy605:
It works for `Player` video type but not for `Default`.

PR #1287, opened 2025-09-14
fix(DefaultVideo): fix loop prop

Marginy605:
Closes https://github.com/gravity-ui/page-constructor/issues/1286

### DIFF
--- a/src/components/DefaultVideo/DefaultVideo.tsx
+++ b/src/components/DefaultVideo/DefaultVideo.tsx
@@ -21,7 +21,13 @@ interface DefaultVideoProps {
 export const DefaultVideo = React.forwardRef<DefaultVideoRefType, DefaultVideoProps>(
     (props, ref) => {
         const {video, qa, customBarControlsClassName} = props;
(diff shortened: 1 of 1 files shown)


src/components/DefaultVideo/DefaultVideo.tsx
```diff
@@ -21,7 +21,13 @@ interface DefaultVideoProps {
 export const DefaultVideo = React.forwardRef<DefaultVideoRefType, DefaultVideoProps>(
     (props, ref) => {
         const {video, qa, customBarControlsClassName} = props;
-        const {controls, customControlsOptions, muted: initiallyMuted = true, onVideoEnd} = video;
+        const {
+            controls,
+            customControlsOptions,
+            muted: initiallyMuted = true,
+            onVideoEnd,
+            loop,
```
NikitaCG, line 29: we won't get a conflict with this logic ? https://github.com/gravity-ui/page-constructor/blob/main/src/components/Media/Video/Video.tsx#L49

 when rendering from the Video file, you need to check that loop is a boolean, and if it is a parameter, then you don't need to include it in defaultVideo

NikitaCG (changes requested):

PR #1319, opened 2025-11-06
chore(ErrorWrapper): Error wrapper story update

KrotovPetr:


src/components/ErrorWrapper/__stories__/ErrorWrapper.stories.tsx
```diff
@@ -5,12 +5,52 @@ import ErrorWrapper, {ErrorWrapperProps} from '../ErrorWrapper';
...
 
 export default {
-    component: ErrorWrapper,
     title: 'Components/ErrorWrapper',
-} as Meta;
+    component: ErrorWrapper,
+    parameters: {
+        layout: 'centered',
+    },
+    argTypes: {
```
aeksandla: This should work automatically. If it doesn't work, then you need to understand why.

Look at the reasons that I described in the comments, maybe it's them?
https://github.com/gravity-ui/page-constructor/pull/1322

PR #1322, opened 2025-11-10
chore(Share): update storybook

lentsd:
Updated Share block story for better storybook ux


src/blocks/Share/__stories__/Share.stories.tsx
```diff
@@ -1,25 +1,43 @@
...
+            options: Object.values(PCShareSocialNetwork),
+        },
+    },
 } as Meta;
 
-uikitConfigure({lang: UIKitLang.En});
-
 const DefaultTemplate: StoryFn<ShareBLockModel> = (args) => (
-    <PageConstructor content={{blocks: [args]}} />
+    <PageConstructor content={{blocks: [blockTransform(args)]}} />
```
aeksandla: Let's remove the PageConstructor here and just make a Share component. Most likely, this will fix the problem with the controls.

lentsd: Removed PageConstructor

PR #1357, opened 2025-12-18
Reafctor(Table, Questions): use JSON-LD for Schema.org definitions.

arteria32:


arteria32:
**Default FAQ Block  from storybook:**

Screenshot 2025-12-18 at 08 25 51

**Default Map from storybook:**
Screenshot 2025-12-18 at 08 19 47

PR #1411, opened 2026-04-07
feat(Hero): add block

PahaN47:


src/blocks/Hero/__stories__/Hero.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+Background.args = {
+    ...data.default,
+    variants: [
+        data.backgroundVariants.color,
+        data.backgroundVariants.fullWidth,
+        data.backgroundVariants.image,
+        data.backgroundVariants.imageFullWidth,
+    ],
+} as VariantsTemplateModel;
+Background.parameters = {
```
aeksandla, line 119: image
There is should be array. For example https://preview.gravity-ui.com/page-constructor/?path=/story/blocks-header--size

src/blocks/Hero/__stories__/Hero.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+export default {
+    title: 'Blocks/Hero',
+    component: Hero,
+    parameters: {
+        controls: {
+            exclude: ['type'],
+        },
+    },
+    argTypes: {
```
aeksandla, line 19: Сontrols should be formed from the prop type

aeksandla: Now it's doesn't work, because file name "Hero" but component name "HeroBlock". File name and component name should be same

src/blocks/Hero/__stories__/Hero.stories.tsx
aeksandla: and in other stories with more then one block should have array in controls

src/blocks/Hero/Hero.scss
```diff
@@ -0,0 +1,237 @@
...
+@import '../../../styles/variables.scss';
+
+$block: '.#{$ns}hero-block';
+
+$backgroundMaxWidth: 1440px;
+$intermediateBreakpoint: 993px;
+
+@mixin media-content-styles($orientation: 'horizontal') {
+    &,
+    .pc-media-component-image__item,
```
aeksandla, line 11: can we use className props instead this?

aeksandla (changes requested):

src/blocks/Hero/Hero.scss
PahaN47: for this we would need to add 5-10 new className props to Media. Would it be worth it?

aeksandla: no, let's leave it as it is

NikitaCG:
what about update memory bank and dependency schemas for hero block ? add in components usage shemas hero block, like this https://github.com/gravity-ui/page-constructor/blob/main/memory-bank/usage/button.md

styles/mixins.scss
```diff
@@ -429,6 +429,10 @@
 @mixin heading1 {
     @include heading(display-4);
 
+    @media (max-width: map-get($gridBreakpoints, 'md')) {
```
NikitaCG, line 432: which blocks still need to be tested after this feature ?

src/blocks/Hero/hooks.ts
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+
+import throttle from 'lodash/throttle';
+
+const SIZE_UPDATE_FREQUENCY_MS = 100;
+
+export const useContainerAspectRatio = () => {
```
NikitaCG, line 7: it is not util, its custom hook

src/blocks/Hero/Hero.scss
```diff
@@ -0,0 +1,237 @@
...
+                @include text-size(body-3);
+
+                color: var(--_--overtitle-color);
+            }
+        }
+    }
+
+    &__media {
+        position: relative;
+        grid-column: span 6;
```
NikitaCG, line 151: 6  - magic number, why six ? what that means

NikitaCG (changes requested):

src/blocks/Hero/Hero.scss
PahaN47: that means that we have a grid of 12 columns and the element takes up 6 of them

aeksandla (approved):

NikitaCG (approved):

PR #1451, opened 2026-05-10
fix(Header): restore background image on mobile devices

SomedayToday:
Fixes [#1805 in `diplodoc-platform/client`](https://github.com/diplodoc-platform/cli/issues/1805) (or the corresponding issue in this repo).

The `renderMedia` condition in `Background` was preventing background images from appearing on mobile because `!isMobile` evaluated to `false`.  
Replaced the condition with `Boolean(imageObject)`, so a background image is shown whenever it's provided, regardless of screen size.

### DIFF
--- a/src/blocks/Header/Header.tsx
+++ b/src/blocks/Header/Header.tsx
@@ -38,7 +38,7 @@ interface BackgroundProps {
 const Background = ({background, isMobile}: BackgroundProps) => {
     const {url, image, fullWidthMedia, video, color} = background;
     const imageObject = url ? getMediaImage(url) : image;
-    const renderMedia = !isMobile || (typeof image === 'object' && 'mobile' in image);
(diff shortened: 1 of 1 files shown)


aeksandla:
The functionality is working correctly now. If you need to put an image on mobile, then you need to use the `mobile` parameter to set the image.
https://preview.gravity-ui.com/page-constructor/?path=/docs/documentation-types--docs#imagedeviceprops

SomedayToday:
@aeksandla thank you for the clarification! I didn't realize the mobile parameter was available for background images via ImageDeviceProps. That makes sense, and the original behavior is indeed correct.

I'll close this PR since it's not needed. Perhaps the documentation for header-block could mention this mobile option more prominently, but that's just a suggestion — the functionality is fine.

Thanks again for reviewing, and sorry for the noise!

PR #1458, opened 2026-05-18
feat: fixed blocks visibility on different screen sizes

separatrixxx:
The `visible` field type has been expanded. It can now accept values ​​of the form `visible: md: true/false` for more flexible display customization


NikitaCG:
Hi! 
1. We want to transform visible prop type from `string` -> `string | {[breakpoint]: boolean}`
2. 
```
{
   sm: false, // dont show on sm breakpoint
   lg: false,
   xl: true, // show on xl
   // if we dont have any breakpoint lg for example, we think is true
}
```

3. In current majoir version we dont want this for children, because they dont have BaseBlock wrapper
4. In next major we are create `visible` in independent wrapper/extension and use it for childrens

about hidden prop - we dont want create more props, it's increases the cognitive load

cc @DaffPunks @aeksandla

src/containers/PageConstructor/components/ConstructorItem/ConstructorItem.tsx
```diff
@@ -32,8 +33,20 @@ export const ConstructorItem = ({
...
+
+    const visibilityClasses = visible ? getVisibleClasses(visible) : '';
+
+    const content = (
+        <BlockDecoration type={data.type as BlockType}>
+            <ConstructorItem data={data} key={data.type} blockKey={blockKey} />
+        </BlockDecoration>
+    );
+
+    if (!visibilityClasses) {
```
aeksandla: maybe should always wrap it in a div for uniformity?

separatrixxx:
It seems very inconvenient to write false for multiple screen sizes. Maybe make one false and not show the block for all sizes larger than that?

aeksandla (changes requested):

aeksandla:
> It seems very inconvenient to write false for multiple screen sizes. Maybe make one false and not show the block for all sizes larger than that?

@separatrixxx 

Hi! I agree with you. It should work as colSizes props. For example {md: false, lg: true}: 
- before `md` block should be visible, 
- between `md` and `lg` block should be hide, 
- after `lg` block should be visible

separatrixxx:
@aeksandla done

aeksandla (approved):

PR #1482, opened 2026-07-14
feat: custom video button theme

aeksandla:


src/components/VideoButton/VideoButton.scss
```diff
@@ -0,0 +1,55 @@
...
+    &_theme_grey {
+        color: var(--pc-text-header-color);
+        background-color: $videoPlayButtonGrey;
+    }
+
+    &_text {
+        padding: $indentXS $indentM;
+        font-weight: 400;
+        font-size: 24px;
+        line-height: 32px;
```
aeksandla, line 57: Copied from the original styles of the button. Replacing it with a mixin is a breaking change.

NikitaCG (commented):
Are we need to update memory-bank or something docs ?

aeksandla:
> Are we need to update memory-bank or something docs ?

updated docs

NikitaCG (approved):

PR #1486, opened 2026-07-21
feat: add React Server Components ('use client') support

KrotovPetr:
- **feat: add use client directive for RSC**
- **feat: add some components to RSC**


package.json
```diff
@@ -29,6 +29,51 @@
       "require": "./server/index.js",
       "import": "./server/index.js"
     },
+    "./models": {
```
NikitaCG, line 32: why we add these imports ?

src/sub-blocks/PriceCard/PriceCard.tsx
```diff
@@ -1,6 +1,11 @@
 import {Check} from '@gravity-ui/icons';
 
-import {BackgroundImage, Buttons, CardBase, ContentList, Links, YFMWrapper} from '../../components';
+import BackgroundImage from '../../components/BackgroundImage/BackgroundImage';
```
NikitaCG, line 3: why we change imports ?

NikitaCG (changes requested):

src/sub-blocks/PriceCard/PriceCard.tsx
KrotovPetr: The common components file exports both server and client components. Direct imports avoid pulling unrelated client components into PriceCard, so it can work as a server component.

package.json
KrotovPetr: > why we add these imports ?

These exports let users import only the files they need. Without them, all imports go through the main entry point, which also includes client components and causes problems with RSC.

